PR #4, opened 2018-06-27
Install fix, example of change to private

ashander:
(using the python convention of single leading underscore => 'private')


petrelharp:
oh, and good point about the subclass!

PR #9, opened 2018-06-30
Allow mutations only at integer sites

petrelharp:
I've written code that - I think - puts down mutations only at integer positions. With this branch, it worksl like this:
```
import msprime
import pyslim

ts = msprime.simulate(10, recombination_rate=0.01, length=100, mutation_rate=0.1)
new_ts = pyslim.mutate(ts, mutation_rate=0.2)

for h in new_ts.haplotypes():
    print(h)
```

It's simulating under an infinite alleles model, to avoid having to check if the parent mutation is the same.  We want this in `tskit` eventually, and implemented in C (this is pure python, so probably slow for big tree sequences) but we could provide this as a stopgap.

### DIFF
--- a/pyslim/__init__.py
+++ b/pyslim/__init__.py
@@ -3,3 +3,4 @@ from __future__ import division
 
 from pyslim.slim_metadata import *       # NOQA
 from pyslim.slim_tree_sequence import *  # NOQA
+from pyslim.mutate import *              # NOQA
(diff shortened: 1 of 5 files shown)


jeromekelleher:
One quick question here: why the ``mutation_rate=0.1`` in the call to ``msprime.simulate()``?

petrelharp:
To check if existing mutations would be retained... which, I don't think they are, correctly.  See https://github.com/tskit-dev/msprime/issues/553

PR #18, opened 2018-09-18
Population annotations are now made correctly.

gtsambos:
Hello,

I noticed that the annotate_defaults() function returns an error when the node tables have non-trivial population info in them.

I'm not super familiar with setting up tests in Python, but I have written a short standalone script (found [here](https://github.com/gtsambos/AdmixtureSim/commit/fecbc66641142bfadc84ec63f44335dccba334d6)) that illustrates the problem.

I think this can be fixed with my commit - feel free to pull it into your repo if you find it useful.

Cheers,
Georgia (PhD student at the University of Melbourne, Australia)


petrelharp:
Gee, thanks for the bug report. I commited a different (more robust?) fix, and a test based on yours.

gtsambos:
Cool, no worries!

PR #26, opened 2018-12-20
Add decoded metadata to objects (and individual time, population)

petrelharp:
Individuals carry a list of nodes, and nodes have `time` and `population` for historical reasons. Here I've propagated these properties to the `individual`, and throw an error if they don't all agree.  Anything else along these lines we should propagate?  Are there any use cases where we expect more than one time or population?

### DIFF
--- a/pyslim/slim_metadata.py
+++ b/pyslim/slim_metadata.py
@@ -39,34 +39,52 @@ class MutationMetadata(object):
     def __ne__(self, other):
         return not self.__eq__(other)
 
+
 def decode_mutation(buff):
     '''
     Extracts the information stored in binary by SLiM in the ``metadata``
     column of a :class:`MutationTable`.  If ``buff`` is empty, returns ``[]``.
+    If `buff` is already a list of MutationMetadata objects, this will return 
+    it, unchanged.
+
+    Note that unlike other metadata decoding functions, this returns a 
+    *list* of MutationMetadata objects, because (thanks to mutation stacking),
+    a given "mutation" as recorded in tskit may actually represent a
+    combination of several SLiM mutations.
 
     :param bytes buff: The ``metadata`` entry of a row of a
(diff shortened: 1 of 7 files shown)


petrelharp:
It could also be nice to provide methods that returns a numpy array of the things in metadata, e.g., ages, population, [location](#28), etc? This can be seen [here](https://github.com/gbradburd/spgr/blob/f4b175a8c51ecedf6196b682e417f60df53b058c/plots/spatial_slim/spatial_slim_ts.py#L268).

petrelharp:
I've added a number of useful things here:

1. Individual objects now have properties that we extract from the metadata.
2. SlimTreeSequences have attributes that are numpy arrays like `ts.individual_ages` and `ts.individual_sex` that act like columns in the individual table.
3. There's methods to ask "who was alive at this time?" and "how old were they at this time?"

The first thing we can copy to allow easier access to metadata for mutations, etcetera.

petrelharp:
At first I tended towards making *all* the properties of metadata also properties of the individual, directly, so you could just type `ind.sex` instead of `ind.metadata.sex`. However, this got awkward for mutations, which have a *list* of metadata objects (thanks to mutation stacking).  So, I've implemented so far `ind.metadata` and `mut.metdata`, e.g.:
```
for ind in ts.individuals():
  print(ind.metadata)
```
Note that these can't be modified (or if they are it won't go back to the tree sequence).

petrelharp:
I would like to merge this; if anyone (@hyanwong? @bhaller? @cjbattey? @jgallowa07? @jeromekelleher? @gtsambos?) has any input on the API, lmk soon.

bhaller:
This sounds reasonable to me, to the extent I understand what's going on in Python.  :->  The "can't be modified" caveat should of course be documented carefully.

petrelharp:
Agreed. I really ought to write some more better documentation...

PR #30, opened 2019-03-04
added simplify

jgallowa07:
So I added the simplify() method that should override the superclass method and return `<class 'pyslim.slim_tree_sequence.SlimTreeSequence'>` after simplifying the tables directly. I'm still confused about the provenance tables; So `The record_provenance` parameter is deprecated until somebody can fix that portion real quick. :)

### DIFF
--- a/pyslim/slim_tree_sequence.py
+++ b/pyslim/slim_tree_sequence.py
@@ -140,6 +140,28 @@ class SlimTreeSequence(tskit.TreeSequence):
         ts = tables.tree_sequence()
         return cls(ts)
 
+    def simplify(self, *args, **kwargs):
+        '''
+        This is a wrapper for msprime.TreeSequence.Simplify().
+        The only difference is that this method returns the
+        derived class SlimTreeSequence.
+
+        Documentation for this function and its arguments
+        can be found at:
+        https://github.com/tskit-dev/tskit/blob/master/python/tskit/trees.py
+
+        :param *args: list of arguments to feed msprime.TreeSequence.Simplify()
+        :param **kwargs: keyword specific aguments stored in a dictionary.
+        :rtype SlimTreeSequence:
+        '''
+        sts = super(SlimTreeSequence,self).simplify(*args, **kwargs)
+        if (type(sts) == tuple):
+            ret = (SlimTreeSequence(sts[0]), sts[1])
+        else:
+            ret = SlimTreeSequence(sts)
+
+        return ret
+
     def recapitate(self, recombination_rate, keep_first_generation=False,
                    population_configurations=None, **kwargs):
         '''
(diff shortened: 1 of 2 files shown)


petrelharp:
Great, thanks!  Let's see:

- Instead of writing out all the parameters, is there any obstacle to using `*args, **kwargs`? Something like:
```
def simplify(*args, **kwargs):
  sts = self.simplify(*args, **kwargs)
  psts = SlimTreeSequence(sts)
  return psts
```
That isn't quite right because of the complications with `map_nodes. But, we should really aim for this being as thin a wrapper around `self.simplify( )` as possible.

- Is all that documentation from `msprime.TreeSequence.simplify`? We should just refer to it, rather than copying it, and only document here any additional changes.

- We need some tests. We can rely on tskit doing simplify right; so here we just need to check that nothing pyslim-ish got messed up.

- I see that `.simplilfy()` already adds a `provenance` entry. The idea with provenance entries to to record where the tree sequence came from. We obviously can't record every single change to a tree sequence, but major ones at least. I don't see what additional changes pyslim will be making that would need to be recorded in the provenance table?

jgallowa07:
Totally agree with all the above. although what you have is recursive. 

how about ```142     def simplify(self,*args,**kwargs):
143         '''
144         wrapper for msprime.TreeSequence.Simplify()
145         '''
146         sts = super(SlimTreeSequence,self).simplify(*args, **kwargs)
147         psts = SlimTreeSequence(sts)
148         return psts```

I believe this should work with the `map_nodes` business too?

petrelharp:
Looks better. But yes, needs to work with `map_nodes`; I am not sure the most pythonic way to do that.

jgallowa07:
oh, I see the problem now. I'll work on this tomorrow.

pyslim/slim_tree_sequence.py
```diff
@@ -139,6 +139,28 @@ def load_tables(cls, tables):
         ts = tables.tree_sequence()
         return cls(ts)
 
+    def simplify(self,*args,**kwargs):
+        '''
+        wrapper for msprime.TreeSequence.Simplify()
+        The only difference is that this returns the
+        derived class SlimTreeSequence
```
petrelharp: complete sentences, please? with capitalization and punctuation?

petrelharp:
Looks good. The only additional attribute that a SlimTreeSequence has is the `.slim_generation`; this should be set properly, but we should check (in a unit test).

petrelharp:
This looks great, @jgallowa07! Thanks a lot. Could you pep8-ify this and squash the commits, then we're good?

jgallowa07:
sure! but fyi I ran `flake8 --max-line-length 89 setup.py pyslim tests` and there are a LOT of things that need fixing to follow pep8, want me to attack these? maybe in a different PR?

jgallowa07:
Squashed -- should be good to go

jgallowa07:
ping @petrelharp can we merge this?

petrelharp:
There was still some non-pep8 here; I fixed this in the `simplify` branch; could you move those changes over here?

petrelharp:
Thanks!!!

PR #37, opened 2019-04-17
add file v0.3 (nucleotides) support

petrelharp:
This puts in the ability to deal with the updated SLiM file version (0.3), which adds `nucleotide` to the MutationMetadata.


petrelharp:
I think this is ready to go. @bhaller, anything in particular we should test that I might not have?

bhaller:
Well, we should try (1) saving from SLiM and loading into pyslim, and (2) generating in msprime, annotating in pyslim, and then loading into SLiM, and (3) simulating in SLiM, importing into pyslim, overlaying mutations with msprime, annotating in pyslim with nucleotide info, and then importing back into SLiM, I suppose.  I'm not sure whether (2) and (3) are part of your plan right now, or if you're just doing (1).

petrelharp:
I've done (1), but have not done (2) or (3) with nucleotides. Good point.

Oh - what about the reference sequence?

bhaller:
Yes, accessing the reference sequence should certainly be part of this base work, even in you don't want to do (2) or (3) right now...

petrelharp:
Hm. I am doing (2) and (3), and how are you recording in the tree sequence file whether it is a nucleotide model or not?

bhaller:
It is not written out explicitly.  In practice, any .trees file written by SLiM from a nucleotide-based model will have an ancestral sequence included, whereas an ancestral sequence will never be present when written from a non-nucleotide-based model.  But the metadata is the same for mutations; a value of -1 will be used when there is no nucleotide associated with a mutation (which can be the case even in nucleotide-based models).  When SLiM loads a .trees file it does various sanity checks; nucleotide values must be -1 for any mutations associated with nucleotide-based mutation types, for example, and must not be -1 for any mutations associated with non-nucleotide-based mutation types.  And an ancestral sequence must be present if loading into a nucleotide-based model (but one may be present when loading into a non-nucleotide-based model, I think, as long as no mutations have nucleotide values other than -1).  So the answer is "apart from sanity/consistency checks, you don't need to know."

petrelharp:
Hey, @bhaller - when you initially proposed using ACGT or 0123 interchangably, it seemed reasonable; but now I see that what's written in the ancestral sequence is ACGT, while in the mutation metadata is 0123. This seems unncecessarily confusing and hard to generalize. Is there a reason we can't have them agree (and I guess I'd vote to have both be ACGT)?

> an ancestral sequence must be present if loading into a nucleotide-based model (but one may be present when loading into a non-nucleotide-based model, I think, as long as no mutations have nucleotide values other than -1)

I haven't got exactly the dependencies of these sanity checks, but the firewall between nucleotide models and non-nucleotide models still seems artificial to me, so I'm hoping to ignore it.

bhaller:
Well, in SLiM internally it is all 0123, for compactness in memory and speed of access and so forth.  I would prefer to minimize the translation back and forth between integers and ASCII; it will only slow down the writing/reading of stuff.  My preference, as I expressed before, would be for the ancestral sequence to be numeric as well, but Jerome wanted to allow non-ACGT characters there I guess, and wanted it to be human-readable, which does make sense.  Anyway, if pyslim wants to make the user-visible nucleotide for mutations be ACGT instead, doing that translation internally, that seems fine.  But I would not be particularly enthusiastic about converting the metadata format to ASCII.

It's fine for pyslim to treat nuc and non-nuc the same, I imagine.  There will then be some things that pyslim could produce that SLiM would refuse to read – mutations with nucleotide values when no ancestral nucleotide sequence is provided, for example – but that's not a big deal.  SLiM will hopefully give a coherent error message in such situations, and the user will fix what they're doing to be compliant.  I don't think pyslim needs to try to duplicate SLiM's sanity-checking, necessarily; that would be an error-prone exercise.

petrelharp:
If we don't make them the same, we need to record what 0123 means somewhere.

bhaller:
Yes, that's in section 24.4.1 as far as the .trees file is concerned.  It's mentioned in lots of other places in the SLiM manual too, since the same values are used in various other contexts.

petrelharp:
> If we don't make them the same, we need to record what 0123 means somewhere.

I meant to record this in the tree sequence. But since we're working with slim-specific metadata, maybe it's ok just to record this as a constant in pyslim:
```
pyslim.NUCLEOTIDE_MAP = ['A', 'C', 'G', 'T']
```
so then nucleotide `k` means `pyslim.NUCLEOTIDE_MAP[k]`.

petrelharp:
Ok, we have a problem here - currently the python interface to `kastore` does not support modifying stores. This means I can't write out the reference sequence to a `.trees` file. Enabling this is going to require some input from @jeromekelleher on extending the `kastore` API. I can probably make it happen if we decide how to do it.

petrelharp:
I'm currently loading the reference sequence (but need to write tests for it), btw.

bhaller:
Yes, Jerome talked about wanting to make .trees files self-annotating, with descriptions of the metadata format they use, but right now we're only documenting the flags column of the individuals table, I think; working out a coherent way to describe the rest of the metadata was punted until some later date.  So just mapping them as you say should be fine (and of course mapping the characters back to integers when going in the opposite direction).

Re: modifying the kastore, yes, makes sense.  We'll see what Jerome says.

jeromekelleher:
> Ok, we have a problem here - currently the python interface to kastore does not support modifying stores. This means I can't write out the reference sequence to a .trees file. Enabling this is going to require some input from @jeromekelleher on extending the kastore API. I can probably make it happen if we decide how to do it.

I can add 'append' mode to the Python kastore interface easily enough, but I'm not sure when I'll be able to get to it as I'm travelling this week. Can you open an issue for this on the kastore repo please @petrelharp? 

Out of interest, why do you need append mode here @petrelharp? I thought SLiM added the ancestral sequence internally.

petrelharp:
> Out of interest, why do you need append mode here @petrelharp? I thought SLiM added the ancestral sequence internally.

We need to do this if we want to modify tree sequences using pyslim, then reload them into SLiM, for instance, if you want to simulate ancestral variation with msprime, then make it non-neutral and use it as a starting point in SLiM.

petrelharp:
I'm testing a bunch of things (including better reloading testing *without* nucleotides) now. I say we merge this, and you go ahead with 3.3, we'll just have the caveat that we can't modify-and-reload tree sequences into SLiM for nucleotide models. We can push that off until we get append into kastore? I have other pyslim homework I'd like to do first.

bhaller:
OK, sure.  SLiM 3.3 is not yet ready to roll, but it's getting there; probably by the end of the month as originally hoped, or slightly after.

bhaller:
So, the SLiM manual ought to have a quick recipe that shows how to access nucleotides and the ancestral sequence in pyslim, perhaps.  Can you throw something together based upon the tests you've been doing?  It can be a trivial model, no need to make it interesting.  Maybe just starting with the recipe named "Modeling elevated CpG mutation rates and equilibrium nucleotide frequencies" (section 17.5 in your draft manual, I think?), and providing a couple of lines of Python showing how to calculate the final nucleotide frequencies in Python with pyslim?  I can write the accompanying text.  Or maybe that example is even too complicated; maybe starting with "A simple neutral nucleotide-based model" (section 17.1 in your manual I think?) and doing a more trivial analysis would suffice.  Anyway, an example would be great.  If you don't have time, feel free to point me toward what I would need and I'll write one myself, I just don't know where to start.  :->

Issue #47, opened 2019-07-31
more intuitive overlap between handling of genetic data and individual metadata


gbradburd:
Hi Peter!

This is a non-urgent request for either new tools, more documentation, or a tutorial for getting genetic data to match up more intuitively to individual data in pyslim.  Some areas of (my) confusion that cropped up in trying to run some nonWF simulations for diploids with tree sequence recording:

1. The individuals included in .genotype_matrix or output by .write_vcf are only the ones flagged as `samples`, while .individual_locations or .individual_times output data for _everyone_. So standard stuff like running a spatial sim and outputting a location and a genotype for each sampled individual requires some hoop-jumpery to output only the metadata for the sample-flagged individual set.

2. The order of the genetic data contained in .genotype_matrix() or output by .write_vcf is determined by the numbering of the nodes associated with the individuals flagged as samples, but the ordering of .individual_locations or .individual_times is determined by individual ID.  This is maybe extra confusing, as the sample names associated with each sample in .write_vcf() aren't informative (ie don't match back to individual metadata).

Hope those are clear - happy to write more about what I was trying to do and where I got mired.  And again - not urgent!  Just wanted to add it to the wishlist...

-g

petrelharp:
Thanks! I'm a little unclear on what exactly the example workflow should be, though. How's this?

1. Simulate data with SLiM
2. Recapitate, mutate, etcetera
3. Take a subsample of the modern individuals
4. Get these individual locations and make a map
5. Compute divergences between individuals using tskit, and plot against geographic distance
6. Write out a VCF file of these individual genotypes for use by other programs

... would that cover things? Want to suggest any other steps?

gbradburd:
Yeah that covers a lot of it.  I'd add (in **bold**):

1. Simulate data with SLiM
2. Recapitate, mutate, etcetera
3. Take a subsample of the modern individuals
**or historical individuals(as remembered in the SLiM model)**
4. Get these individual locations and make a map
5. Compute divergences between individuals using tskit, and plot against geographic distance
**and write out divergences for use by other programs**
6. Write out a VCF file of these individual genotypes for use by other programs
7. **Write out locations/times/metadata of these individuals for use by other programs in a way that makes it easy to match up to genetic data (divergences or genotypes) to metadata, either because they have the same order, or because both sets of output can be associated with individual IDs**

Does that make more sense?  Thanks Peter!

petrelharp:
This got done in #48, I think.

PR #48, opened 2019-08-06
vignette!

petrelharp:
Here's a draft of the answer to @gbradburd's nice suggestion in #47.  It'd be great if some other folks could read this over and give suggestions or make edits. In particular,

1. I'm sure there needs to be more explanation in some places of what's going on. Which places are those?
2. The spatial simulation could probably be tweaked to make a nicer demonstration?

Any input, @gbradburd, @bhaller, @cjbattey?  To view the html version of this, you should be able to do `make` in the `docs/` directory.


docs/vignette_space.rst
```diff
@@ -0,0 +1,584 @@
...
+   We have 35 individuals in the bottomright group.
+   We have 46 individuals in the center group.
+   We have 5 individuals in the ancient group.
+
+
+To keep names associated with each subset of individuals,
+we've kept the individuals in a dict, so that for instance
+`corners["topleft"]` is an array of all the individual IDs that are in the top left corner.
+The IDs of the ancient individuals we will work with are kept in the array `ancient`.
+
```
petrelharp: Hey look you can make notes at specific locations. Like this one: here would be a good place to double-check that all the `ancient` samples are from a while ago, like:
```
for i in corners["ancient"]:
   ind = ts.individual(i)
   assert(ind.time >= 500 and ind.time < 520)
```

docs/vignette_space.rst
```diff
@@ -0,0 +1,584 @@
...
+   import msprime
+
+   recap_ts = slim_ts.recapitate(recombination_rate=1e-8, Ne=2000)
+   ts = pyslim.SlimTreeSequence(
+         msprime.mutate(recap_ts, rate=1e-8, keep=True))
+
+   print(f"The tree sequence now has {ts.num_trees} trees,"
+         f" and {ts.num_mutations} mutations.")
+
+Note since :meth:`mutate <msprime.mutate>` is an msprime method, it does not return a pyslim
```
jeromekelleher: Could use a `.. note`` or ``.. warning`` block here to make this stand out.

docs/vignette_space.rst
```diff
@@ -0,0 +1,584 @@
...
+
+
+******************
+Plotting locations
+******************
+
+We should check this: plot where these individuals lie
+relative to everyone else.
+The individuals locations are available in individual metadata,
+but to make things easier, it's also present in a `num_individuals x 2`
```
jeromekelleher: looks like N x 3 below?

jeromekelleher:
I had a read through @petrelharp, looks excellent to me. This will really help people out. 

We've got to get VCF and metadata sorted out soon in tskit, the last bit really is unnecessarily awkward and confusing.

petrelharp:
> We've got to get VCF and metadata sorted out soon in tskit, the last bit really is unnecessarily awkward and confusing.

omg, yes.

cjbattey:
looks great - very helpful. It may be useful to add a note that the rates in recapitate( ) and mutate( ) are per-timestep and not per-generation. This is also relevant around line 230-235 where you discuss keeping SLiM mutations, because the rates in SLiM and the rates applied by mutate( ) will be in different units. Also very much yes on the VCF output.

docs/vignette_space.rst
```diff
@@ -0,0 +1,584 @@
...
+Now let's plot genetic distance against geographic distance.
+
+.. code-block:: python
+
+   pair_colors = np.repeat(0, len(pairs))
+   for k, (i, j) in enumerate(pairs):
+      if ind_group[i] == "ancient" or ind_group[j] == "ancient":
+         pair_colors[k] = 1
+
+   fig = plt.figure(figsize=(6, 6))
```
cjbattey:
```suggestion
   fig = plt.figure(figsize=(6, 6),dpi=300)
```
will make things a little prettier.

gbradburd:
This looks great and I think will be super useful, especially to me!  Agreed that the last chunk is a little complicated, so that it'd be nice to have some of it taken care of by write_vcf(), but I recognize that that's complicated. Thanks for putting the vignette together, and feel free to close https://github.com/tskit-dev/pyslim/issues/47 as you see fit.

PR #57, opened 2019-11-25
Vignette edits

lclclclclclclc:
hi peter et al :)  hope all's well

I ran through the spatial vignette today and noticed a couple errors.


petrelharp:
Hi, Elsie! Thanks a bunch for this; looks good. I will double-check, then ask you to squash these commits down to one? (with `git rebase -i origin/master`)

petrelharp:
Hm, I see the code output isn't sync'ed up either, eg:
Screenshot from 2019-11-25 15-42-18

I think I know what happened...

petrelharp:
Hello! So, the output needs to by sync'ed to the code; I must have changed some parameters and updated the output but not the code, or vice-versa. (I was trying to be careful about that...)  Would you be willing to run the code and make the output agree? (Note that it sets a seed, so should be reproducible.)

lclclclclclclc:
Hi there!

Yes— I will certainly run through it again. 
> and make the output agree?
> 
To check I understand:  I will run through the code as provided and edit the text of the vignette to match the code output? 

That’s what I attempted to do with the pull request. (although I’m worried I messed something up with the squashing..)

I’m happy to run through again and doublecheck though. 

hope all’s well,
elsie

petrelharp:
> To check I understand: I will run through the code as provided and edit the text of the vignette to match the code output?

Yes!  Great.

> That’s what I attempted to do with the pull request. (although I’m worried I messed something up with the squashing..)

Well, when I run the first code chunk, I get
```
The tree sequence has 27151 trees on a genome of length 100000000.0, 3777 individuals, 6304 'sample' genomes, and 0 mutations.
```
whereas the vignette says the output says e.g. `43620 trees` - hopefully, you get the same thing as me?

petrelharp:
This PR is superceded by #69, I think.

PR #79, opened 2020-05-01
parsing provenance

petrelharp:
This might be useful to you, @mufernando - could you have a look and see if it does things you'd want to do?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,10 @@
 
 - added has_individual_parents, a method to find individuals with all parents
   are also recorded as individuals
-- added the `.slim_provenances` property to return all SLiM provenance entries
+- Provenance handling:
+   * added the `.slim_provenances` property to return all SLiM provenance entries
+   * added the `slim_provenance_version` and `parse_provenance` methods to tell if
+      provenance entries come from SLiM and to parse them
 
 ********************
 [0.401] - 2020-03-27
--- a/pyslim/provenance.py
+++ b/pyslim/provenance.py
@@ -19,77 +19,17 @@ class ProvenanceMetadata(object):
     file_version = attr.ib()
 
 
-def get_provenance(ts, only_last=True):
-    '''
-    Extracts model type, slim generation, and remembmered node count from either
-    the last entry in the provenance table that is tagged with "program"="SLiM"
-    (if ``only_last=True``) or a list of all of them (otherwise).
-
-    :param SlimTreeSequence ts: The tree sequence.
-    :param bool only_last: Whether to return only the last SLiM provenance entry,
-        (otherwise, returns a list of all SLiM entries).
-    :rtype ProvenanceMetadata:
-    '''
-    provenances = []
-    for j, p in enumerate(ts.tables.provenances):
-        this_record = json.loads(p.record)
-        is_slim, this_file_version = _slim_provenance_version(this_record) 
-        if is_slim:
-            record = this_record
-            file_version = this_file_version
-            if file_version == "0.1":
-                out = ProvenanceMetadata(record['model_type'],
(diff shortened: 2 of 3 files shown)


mufernando:
It would be good to return the SLiM command and seed.

petrelharp:
> It would be good to return the SLiM command and seed.

Ah, so add these to the `pyslim.Provenance` structure?  Or maybe just include the whole parsed provenance dictionary, really?

mufernando:
With your  second option do you mean parsing all the different fields in the JSON and including them all?

I think most users wouldn't need more than the seed and the SLiM command, but don't see a huge downside to parsing the whole thing.

Let me know and I can help implementing either of these.

petrelharp:
Ok, let's see. Right now, we can get information by parsing the provenance entry with `json` directly:
```
prov = json.loads(ts.provenance(0).record)
for k in prov:
  if isinstance(prov[k], dict):
    print(k, ":")
    for kk in prov[k]:
        print("   ", kk, prov[k][kk])
  else:
    print(k, prov[k])
```
which gets
```
environment :
    os {'machine': 'x86_64', 'node': 'skua', 'release': '4.19.0-2-amd64', 'system': 'Linux', 'version': '#1 SMP Debian 4.19.16-1 (2019-01-17)'}
metadata :
    individuals {'flags': {'16': {'description': 'the individual was alive at the time the file was written', 'name': 'SLIM_TSK_INDIVIDUAL_ALIVE'}, '17': {'description': 'the individual was requested by the user to be remembered', 'name': 'SLIM_TSK_INDIVIDUAL_REMEMBERED'}, '18': {'description': 'the individual was in the first generation of a new population', 'name': 'SLIM_TSK_INDIVIDUAL_FIRST_GEN'}}}
parameters :
    command ['slim', 'recipe_WF.slim']
    model initialize()
{
    setSeed(23);
    initializeTreeSeq();
    initializeMutationRate(1e-2);
    initializeMutationType("m1", 0.5, "f", -0.1);
    initializeGenomicElementType("g1", m1, 1.0);
    initializeGenomicElement(g1, 0, 99);
    initializeRecombinationRate(1e-2);
}

1 { 
    sim.addSubpop("p1", 10);
}

10 {
    sim.treeSeqOutput("recipe_WF.trees");
    catn("Done.");
    sim.simulationFinished();
}

    model_hash c97711cae1feb5f86fe9d948c34167bbdae1d8fe895a2a14da517a24425a9f04
    model_type WF
    seed 1711090956800
schema_version 1.0.0
slim :
    file_version 0.4
    generation 10
software :
    name SLiM
    version 3.3.2
```

And, from the pre-parsed information:
```
ts.slim_provenance
# ProvenanceMetadata(model_type='WF', slim_generation=10, file_version='0.4')
```
I don't think we want to parse *everything* here - for instance, one of the entries is `metadata`, which tells us what the individual flags that SLiM is using mean; that's internal to pyslim. I was suggesting adding another slot, say `record` to the `ProvenanceMetadata` class, that has the dictionary returned by `json.loads(ts.provenance(j).record)` in it (ie the first thing).  Then one could get `seed` by doing
```
ts.slim_provenance.record['parameters']['seed']
```

This would make everything accessible, which is good - although it is already accessible with just a `json.loads( )`. If there's just a few things that are obviously and often useful then we should add them as properties of the provenance metadata itself. I'm not sure if `seed` is one of those things, since it seems like we should put things in the `ProvenanceMetadata` class things that we might want to use within pyslim itself?  Thoughts?

mufernando:
Yes, I think just dumping the whole record onto the `ProvenanceMetadata` class is the best thing. And then people can parse it to their liking.

petrelharp:
I'm going to merge this and we can extend the provenance metadata separately.

PR #89, opened 2020-08-11
Update tskit

petrelharp:
This goes along with https://github.com/MesserLab/SLiM/pull/101 and supercedes #82.


pyslim/slim_metadata.py
```diff
@@ -11,6 +13,431 @@
...
+            "properties": {
+                "slim_id": {
+                    "type": "integer",
+                    "description": "The 'pedigree ID' of this chromosome in SLiM.",
+                    "binaryFormat": "q",
+                    "index": 0},
+                "is_null": {
+                    "type": "integer",
+                    "description": "Whether this node describes a 'null' (non-existant) chromosome.",
+                    "binaryFormat": "B",
```
benjeffery, line 159: How come this isn't a `?` (`bool`). The example below suggests it is.

pyslim/slim_metadata.py
```diff
@@ -11,6 +13,431 @@
...
+                    "description": "The ID of the subpopulation the individual was part of, either when the tree sequence was written out (if the individual was alive then), or the last time they were Remembered (if not).",
+                    "binaryFormat": "i",
+                    "index": 3
+                },
+                "sex": {
+                    "type": "integer",
+                    "description": "The sex of the individual (0 for female, 1 for male, -1 for hermaphrodite).",
+                    "binaryFormat": "i",
+                    "index": 4
+                },
```
benjeffery, line 207: This could also have `"enum": [-1,0,-1]` added.

pyslim/slim_metadata.py
```diff
@@ -11,6 +13,431 @@
...
+                            "migration_rate": {
+                                "type": "number",
+                                "description": "The fraction of children in this subpopulation that are composed of 'migrants' from the source subpopulation (in WF models).",
+                                "binaryFormat": "d",
+                                "index": 2}
+                        },
+                        "required": ["source_subpop", "migration_rate"],
+                        "additionalProperties": False
+                    }
+                }
```
benjeffery, line 315: How is the length of this array encoded? Is it relying on the default array length type being `Q`? Might be nice to be explicit about that?

pyslim/slim_metadata.py
```diff
@@ -11,6 +13,431 @@
...
+
+def set_tree_sequence_metadata(tables,
+        model_type,
+        slim_generation,
+        spatial_dimensionality='',
+        spatial_periodicity='',
+        separate_sexes=False,
+        nucleotide_based=False):
+    if tables.metadata_schema.schema is None:
+        if len(tables.metadata) > 0:
```
benjeffery, line 407: Probably best here to use `isinstance(tables.metadata, bytes)` to test for raw metadata.

pyslim/slim_metadata.py
```diff
@@ -11,6 +13,431 @@
...
+            raise ValueError("Tree sequence has top-level metadata but no schema: this is a problem.")
```
benjeffery, line 408: Maybe "this is a problem as pyslim is attempting to add to the metadata?"

pyslim/slim_metadata.py
petrelharp: That's because in [the SLiM code](https://github.com/petrelharp/SLiM/blob/tskit_update/core/slim_sim.h#L135) they're defined as `uint8_t`. Maybe this would all work out fine and be equivalent? I don't know?

petrelharp: but, they *are* boolean, yes.

pyslim/slim_metadata.py
petrelharp: Right - but, some arguments against: (a) we'd be causing ourselves future headaches if we wanted to add to the list; and (b) perhaps it would slow down decoding? But I haven't tested to see how much it would.

pyslim/slim_metadata.py
petrelharp: [Here's](https://github.com/tskit-dev/pyslim/blob/master/pyslim/slim_metadata.py#L399) a handy reference. And, [here's](https://github.com/tskit-dev/pyslim/blob/master/pyslim/slim_metadata.py#L465) how I was doing the decoding. (I can get rid of that code now, though!) Oh, I see - I just need to put an `"arrayLengthFormat" : "Q"` entry in here, huh?

petrelharp: oh, weird - it looks like I should have done `"arrayLengthFormat" : "i"` - how did this even work?

petrelharp: uh-oh, but the documentation says that `i` is not an option, only `I`!

pyslim/slim_metadata.py
petrelharp: Hm, whoops - I abandoned the idea of SLiM trying to modify the metadata in place. I suppose I could maintain the ability of pyslim to do this...

pyslim/slim_metadata.py
petrelharp: I'll do `"arrayLengthFormat" : "I"` - https://github.com/tskit-dev/tskit/issues/769

petrelharp: ... and, we've changed the specification in C++ to `uint32_t`, since @bhaller says that's OK. =)

pyslim/slim_metadata.py
benjeffery: I think `bool` size is compiler-dependant in C++ so SLiM needs to be specify `uint8_t`. In python `?` will always be one-byte. Should all work out fine.

pyslim/slim_metadata.py
benjeffery: Would slightly slow down validation - but not encoding or decoding which will ignore it.

pyslim/slim_metadata.py
benjeffery: Phew, glad that worked out - sorry that the docs had the default wrong - will fix.

pyslim/slim_metadata.py
petrelharp: Ok - so I'll leave it as-is (right?).

pyslim/slim_metadata.py
petrelharp: Hm. I think we'll omit it, since it's documented here, which seems good enough, to avoid future headaches.

pyslim/slim_metadata.py
bhaller: I think @benjeffery was saying we need to use `?` in the schema, and make sure that the corresponding value in SLiM is always a `uint8_t`; we should never used `b` or `B` in the schema, and we should never use `bool` in the corresponding C++ code.  Do I have that right?

benjeffery: Yes, ? In python uint_8 in C++. Sorry I wasn't clearer!

petrelharp: Oh, I got it. =)

petrelharp: fixed in 151875a698

PR #94, opened 2020-08-28
Use only public tskit API

benjeffery:
Here's a quick sketch of how it could work without subclassing `tskit.TreeSequence`. There's a bit more boilerplate, so it is a trade-off.

### DIFF
--- a/pyslim/slim_tree_sequence.py
+++ b/pyslim/slim_tree_sequence.py
@@ -21,6 +21,27 @@ INDIVIDUAL_FIRST_GEN = 2**18
 # something that in reference_sequence is NUCLEOTIDES[k]
 NUCLEOTIDES = ['A', 'C', 'G', 'T']
 
+class SimpleContainerSequence:
+    """
+    Simple wrapper to allow arrays of SimpleContainers (e.g. edges, nodes) that have a
+    function allowing access by index (e.g. ts.edge(i), ts.node(i)) to be treated as a
+    python sequence, allowing forward and reverse iteration.
+    """
+
+    def __init__(self, getter, length):
+        self.getter = getter
+        self.length = length
+
+    def __len__(self):
+        return self.length
+
+    def __getitem__(self, index):
+        if index < 0:
+            index += len(self)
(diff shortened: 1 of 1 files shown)


petrelharp:
Ah, I see! Hm, thanks. It's not too much more boilerplate. What do you think? Is this a better way to do things?

petrelharp:
Hm, to use only the public API I'll need to figure out a way around this, too: https://github.com/tskit-dev/pyslim/blob/update_tskit/pyslim/slim_tree_sequence.py#L380

petrelharp:
Gee, this has the downside that tab completion no longer works. But it seems to be solving some other problems, so...

petrelharp:
Hmph, another annoyance is that `isinstance(ts, tskit.TreeSequence)` is no longer true.

petrelharp:
And, `@property` appears to not work either.

(Apologies for my stream-of-consciousness notes here...)

jeromekelleher:
I was wondering if the SlimTreeSequence class is actually needed at all now - other than the ``recaptitate`` method, it's not really doing much that tskit isn't now, right?

The ``recapitate`` method could be made a function instead and monkey-patched in to the TreeSequence object for backwards compatibility.

petrelharp:
That's true - but, it does have a few extra properties: `.slim_generation` and `.reference_sequence` (sometimes). Although the first is really part of top-level metadata now.

I've pulled these commits into #95, btw.

benjeffery:
I'll close this PR if you've used the commits in #95. Thanks!

petrelharp:
Gah, there's a significant problem with this approach: anywhere that code uses `isinstance(ts, TreeSequence)` will fail. In the msprime code this happens [here](https://github.com/tskit-dev/msprime/blob/master/msprime/ancestry.py#L314) and [here](https://github.com/tskit-dev/msprime/blob/master/msprime/provenance.py#L128). The second one (provenance construction) means I need to provide an `asdict` method so that, for instance, `msprime.mutate( )` will work on a SlimTreeSequence. I'm not sure how to get around the first one. But this is seeming fragile.

PR #95, opened 2020-08-28
remove first gen

petrelharp:
... and make use of 'stage' in metadata. Depends on #89.


petrelharp:
Ready to go - passing tests with dev msprime and 0.7.5!

PR #97, opened 2020-09-07
Deprecate old metadata handling

### LINKED ISSUE #96: actually deprecate old metadata structures
Although pyslim uses the new metadata stuff, it serves up metadata as structures (eg NodeMetadata objects) rather than just using the dicts that tskit gives us. This is only for backwards compatibility - all the same information is there and the UI is minimally different - and it'd be real nice to GET RID of all this now-superfluous code. But I hadn't thought of a good way to do this.

One important reason to shift to being compatible is that the conflict in notation between tskit and pyslim is super confusing: right now if ts is a TreeSequence you do `node.metadata['slim_id']`, whereas if its a SlimTreeSequence you do `node.metadata.slim_id`.

But now, I think I have a good way to deprecate the old stuff. I could add a `legacy_metadata=False` argument to `pyslim.load( )`, that if `True` would return the old-style metadata, and be removed at some point in the future. This would mean that existing scripts that use metadata would break, but only require adding `legacy_metadata=True` to any calls to `load()` to keep working. (Or, better, going through and changing `node.metadata.slim_id` to `node.metadata['slim_id']`.)

How's this sound?

petrelharp:
Closes #96.

Old scripts should still entirely work as long as `pyslim.load(...)` is replaced with `pyslim.load(..., legacy_metadata=True)`. All the old metadata handling raises DeprecationWarnings.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -5,7 +5,7 @@ Development
 ===========
 
 
-To install a particular version of `pyslim` from source, e.g., to obtain a recent update::
+To install a particular version of ``pyslim`` from source, e.g., to obtain a recent update::
 
 
    $ git clone https://github.com/tskit-dev/pyslim.git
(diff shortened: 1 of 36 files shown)


petrelharp:
I think this is ready to go, except for a way to make the old style throw an informative error...

petrelharp:
Ok, now with some python magic if someone tries to use the old metadata methods (e.g. `node.metadata.genome_type` instead of `node.metadata["genome_type"]`) they'll get an informative error.

benjeffery:
Really like what you've done with the informative error - I think given the ease of upgrade this is justified.

petrelharp:
Whew! I thought for sure you'd suggest a much more elegant way to do it.

PR #101, opened 2020-09-14
switch over to msprime 1.0 API

petrelharp:
Here's a start at that.

To view these docs, you can
```
git fetch origin pull/101/head:msprime_v1
git checkout msprime_v1
cd docs
make
```
and then direct a web browser to the `docs/_build/html` directory.


jeromekelleher:
Should we be using ``sim_ancestry`` rather than ``simulate``?

petrelharp:
> Should we be using sim_ancestry rather than simulate?

Yes! I'm just waiting 'til the new API has settled down to switch everything over. (has it settled, then?)

jeromekelleher:
> Yes! I'm just waiting 'til the new API has settled down to switch everything over. (has it settled, then?)

No, but soon. Next job after I land this big tables refactoring.

petrelharp:
Mostly superceded by #154.

PR #112, opened 2020-12-08
put back ability to set slim_generation

### LINKED ISSUE #110: slim_generation is now a property, not a class variable, so can't be assigned to
Previously, SlimTreeSequence.slim_generation was just a class variable, and thus mutable. Indeed, [stdpopsim modifies it](https://github.com/popsim-consortium/stdpopsim/blob/642c668f39e3115f1b07a0a274f980a7554fb33e/stdpopsim/slim_engine.py#L1212). However, now `slim_generation` is coming from [top-level metadata](https://github.com/tskit-dev/pyslim/blob/70a00baa9eb2f951fecf0a15f165058d9238aad1/pyslim/slim_tree_sequence.py#L193), and so I made it a @property, since otherwise it would not be in sync with the metadata. However, the tree sequence (and its metadata) are immutable, so I can't make a corresponding @setter.

The usage in stdpopsim is easily fixed - we just need to modify the metadata a few lines earlier, when it's still a table collection and not yet a tree sequence. But, this is a breaking change, and I can't see a way around it. Any other ideas...  @benjeffery?

I guess I could at least define a @setter that throws an appropriate error and explains what to do?

 (also see https://github.com/popsim-consortium/stdpopsim/pull/706 )

petrelharp:
Closes #110, and it's terrible, but should work. Perhaps you can have a look, @grahamgower?

### DIFF
--- a/pyslim/slim_tree_sequence.py
+++ b/pyslim/slim_tree_sequence.py
@@ -166,6 +166,8 @@ class SlimTreeSequence(tskit.TreeSequence):
             ts = tables.tree_sequence()
         super().__init__(ts._ll_tree_sequence)
         self.reference_sequence = reference_sequence
+        # need this for backwards compatibility
+        self._slim_generation = ts.metadata['SLiM']['generation']
         # pre-extract individual metadata
         self.individual_locations = ts.tables.individuals.location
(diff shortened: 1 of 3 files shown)


grahamgower (approved):
LGTM!

PR #118, opened 2020-12-21
Continue simulation

jelliotthiggins:
Hello! 

In using pyslim, I found the process described in "Vignette: Following up with more coalescent simulation" incredibly useful. To make it more accessible, I added it as a SlimTreeSequence method called `continue_simulation`. It takes in the time to simulate in front of the slim simulation, as well as msprime args, and outputs a complete tree sequence. It uses one helper method, `adjust_tables_time`, which performs the time adjustment on the tables - I put this in its own method as it could be useful elsewhere. Much of this code came directly from the example, and I tried to maintain consistency in the docstrings and style. I used `recapitate` as a reference for parts of the docstrings and for checking parameters. Below I'll demonstrate how this would simplify the example found in the tutorial. The ability to combine slim simulations with msprime simulations on either end, using this technique and recapitation, is so powerful and I hope this tool can be of use to others. Let me know if you see any issues!

All the best,
Jack Elliott-Higgins

```
import pyslim, msprime

ts = pyslim.load("rapid_adaptation.trees")
rts = ts.recapitate(Ne=1000, recombination_rate=1e-8, random_seed=6)
rts = pyslim.SlimTreeSequence(msprime.mutate(rts, rate=1e-8, random_seed=7))
full_ts = rts.continue_simulation(1000, Ne=10000, samples=20000, mutation_rate=1e-8, recombination_rate=1e-8, random_seed=9)
```

### DIFF
--- a/pyslim/slim_tree_sequence.py
+++ b/pyslim/slim_tree_sequence.py
@@ -490,6 +490,96 @@ class SlimTreeSequence(tskit.TreeSequence):
             recap = tables.tree_sequence()
         return SlimTreeSequence(recap, reference_sequence=self.reference_sequence)
 
+    def resume_simulation(self, time, Ne, samples, 
+                            mutation_rate=None, 
+                            recombination_rate=None,
+                            **kwargs):
+        """
+        some explanation here
+
+        :param int time: The desired period of time to simulate in msprime,
+            in units of generations
+        :param int Ne: The effective population size for :meth:`msprime.simulate`,
+            in units of diploid individuals
+        :param int samples: Sample count for msprime.simulate,
+            in units of haploid samples,
+            should be twice Ne used for SLiM simulation
+        :param float mutation_rate: A (constant) mutation rate,
+            in units of mutations per nucleotide per unit of time.
+        :param float recombination_rate: A (constant) recombination rate,
+            in units of crossovers per nucleotide per unit of time.
+        :param dict kwargs: Any other arguments to :meth:`msprime.simulate`.
+        """
+        if "keep_first_generation" in kwargs:
+            raise ValueError("The keep_first_generation argument is deprecated:"
+                             "the FIRST_GEN flag is no longer used.")
+
+        # toggle for hacks below to deal with old msprime
+        discrete_msprime = hasattr(msprime, "RateMap")
+
+        if recombination_rate is not None:
+            if recombination_map is not None:
+                raise ValueError("Cannot specify length/recombination_rate along with a recombination map")
+            if discrete_msprime:
+                recombination_map = msprime.RecombinationMap(positions = [0.0, self.sequence_length],
+                                                             rates = [recombination_rate, 0.0])
+            else:
+                recombination_map = msprime.RecombinationMap(positions = [0.0, self.sequence_length],
+                                                             rates = [recombination_rate, 0.0],
+                                                             num_loci = int(self.sequence_length))
+        if discrete_msprime and ('discrete_genome' not in kwargs):
+            kwargs['discrete_genome'] = True
+
+        if population_configurations is None:
+            population_configurations = [msprime.PopulationConfiguration()
+                                         for _ in range(self.num_populations)] 
+        new_ts = msprime.simulate(
+                            samples,
+                            Ne=Ne,
+                            end_time=time,
+                            length=self.sequence_length,
+                            mutation_rate=mutation_rate,
+                            population_configurations = population_configurations,
+                            recombination_map = recombination_map
+                            **kwargs)
+
+        new_nodes = np.where(new_ts.tables.nodes.time == slim_time)[0]
+        slim_indivs = self.individuals_alive_at(0)
+        slim_nodes = []
+        for ind in slim_indivs:
+            slim_nodes.extend(self.individual(ind).nodes)
+        slim_nodes = np.array(slim_nodes)
+        assert(len(slim_nodes) == samples)
+
+        node_map = np.repeat(tskit.NULL, new_ts.num_nodes)
+        node_map[new_nodes] = np.random.choice(slim_nodes, len(new_nodes), replace=False)
+
+        tables = self.tables
+        tables.nodes.set_columns(
+                flags=self.tables.nodes.flags & ~np.uint32(tskit.NODE_IS_SAMPLE),
+                time=self.tables.nodes.time + time,
+                population=self.tables.nodes.population,
+                individual=self.tables.nodes.individual,
+                metadata=self.tables.nodes.metadata,
+                metadata_offset=self.tables.nodes.metadata_offset,
+                metadata_schema=json.dumps(self.tables.nodes.metadata_schema.schema))
+        tables.mutations.set_columns(
+                site=self.tables.mutations.site,
+                node=self.tables.mutations.node,
+                time=self.tables.mutations.time + time,
+                derived_state=self.tables.mutations.derived_state,
+                derived_state_offset=self.tables.mutations.derived_state_offset,
+                parent=self.tables.mutations.parent,
+                metadata=self.tables.mutations.metadata,
+                metadata_offset=self.tables.mutations.metadata_offset,
+                metadata_schema=json.dumps(self.tables.mutations.metadata_schema.schema))
+
+        tables.union(new_ts.tables, node_map,
+                add_populations=False,
+                check_shared_equality=False)
+
+        return pyslim.SlimTreeSequence(tables.tree_sequence())
+
     def mutation_at(self, node, position, time=None):
         '''
         Finds the mutation present in the genome of ``node`` at ``position``,
--- a/pyslim/slim_tree_sequence.py
+++ b/pyslim/slim_tree_sequence.py
@@ -490,7 +490,8 @@ class SlimTreeSequence(tskit.TreeSequence):
             recap = tables.tree_sequence()
         return SlimTreeSequence(recap, reference_sequence=self.reference_sequence)
 
-    def continue_simulation(self, slim_time, Ne, 
+    def continue_simulation(self, slim_time, 
+                            Ne=None, 
                             samples=None,
                             mutation_rate=None, 
                             recombination_rate=None,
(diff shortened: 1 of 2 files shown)


petrelharp:
Hello, and welcome! This is an excellent idea, and thanks very much for jumping into it! I've had a read through and have some minor comments that I'll make on the code. But, my first thought is: should this perhaps be an msprime method? I think this isn't doing anything SLiM-specific, really, it's just adding some more coalescent simulation on the end of a tree sequence? What do you think? If so, I can assist or advise on moving the code over there.

pyslim/slim_tree_sequence.py
```diff
@@ -490,6 +490,139 @@ def recapitate(self,
             recap = tables.tree_sequence()
         return SlimTreeSequence(recap, reference_sequence=self.reference_sequence)
 
+    def continue_simulation(self, slim_time, 
+                            Ne=None,
```
petrelharp, line 494: I think you can leave out `Ne`, as we don't do anything with it, so it'll be carried along with **kwargs?

pyslim/slim_tree_sequence.py
```diff
@@ -490,6 +490,139 @@ def recapitate(self,
...
+
+        slim_indivs = self.individuals_alive_at(0)
+        slim_nodes = []
+        for ind in slim_indivs:
+            slim_nodes.extend(self.individual(ind).nodes)
+        slim_nodes = np.array(slim_nodes)
+    
+        if samples is None:
+            samples=len(slim_nodes)
+        assert(len(slim_nodes) == samples)
```
petrelharp, line 565: I think you meant
```
if samples is None:
    samples = np.arange(len(slim_nodes))
```

Also, I don't think there's any need for `assert len(slim_nodes) == samples`? We could continue the simulation and only be interested in a small number of lineages, right? Or am I missing something?

pyslim/slim_tree_sequence.py
```diff
@@ -490,6 +490,139 @@ def recapitate(self,
...
+                recombination_map = msprime.RecombinationMap(positions = [0.0, self.sequence_length],
+                                                             rates = [recombination_rate, 0.0])
+            else:
+                recombination_map = msprime.RecombinationMap(positions = [0.0, self.sequence_length],
+                                                             rates = [recombination_rate, 0.0],
+                                                             num_loci = int(self.sequence_length))
+        if discrete_msprime and ('discrete_genome' not in kwargs):
+            kwargs['discrete_genome'] = True
+
+        slim_indivs = self.individuals_alive_at(0)
```
petrelharp, line 557: hm, if this becomes an msprime method we'll have to include another argument that says which nodes are the ones "alive at the end" (which could default to the samples)

pyslim/slim_tree_sequence.py
```diff
@@ -490,6 +490,139 @@ def recapitate(self,
...
+
+        tables = self.tables
+        self.adjust_tables_time(tables, slim_time)
+        tables.union(new_ts.tables, node_map,
+                add_populations=False,
+                check_shared_equality=False)
+
+        return SlimTreeSequence(tables.tree_sequence())
+
+    def adjust_tables_time(self, tables, time):
```
petrelharp, line 591: Hm - I think *this* should actually be a `tskit` method of a TableCollection (confusingly), since we've wanted to do it in a few places separately now. In the interest of not doing too many things at once, it might be better to just leave it here as a 'private' method by putting an underscore at the start of the name (but as `pyslim._adjust_tables_time`, not as a method of a tree sequence).

pyslim/slim_tree_sequence.py
jelliotthiggins: Definitely agree with this.

pyslim/slim_tree_sequence.py
jelliotthiggins: To the best of my understanding, the number of individuals alive at time=0 in the slim tree sequence must be equal to the sample size simulated in msprime, as we map the two sets of nodes together. I may be missing something though! Or alternatively we could map msprime lineages to arbitrary slim individuals. Not sure how much control to exert over this within the method, but for those two sizes to be different would require a more general use way of linking nodes I believe. 

This is derived from the following lines from the tutorial - i think the variable samples should be renamed sample_size though, a bit confusing there.
```
slim_indivs = rts.individuals_alive_at(0)
slim_nodes = []
for ind in slim_indivs:
   slim_nodes.extend(ts.individual(ind).nodes
slim_nodes = np.array(slim_nodes)
assert(len(slim_nodes) == 20000)
```

pyslim/slim_tree_sequence.py
jelliotthiggins: Yes this may be the only pyslim specific line. Wouldn't be too tough to implement though!

pyslim/slim_tree_sequence.py
jelliotthiggins: That works for now!

jelliotthiggins:
Hi Dr. Ralph! Thanks so much for the feedback - I think you're right that it is more useful as an msprime method and it could always be called within a pyslim wrapper to preserve specific functionality, like is done with `msprime.mutate()`. I would appreciate any assistance or advice moving it there! I'll make those changes and add some more test coverage today and we can go from there.

pyslim/slim_tree_sequence.py
jelliotthiggins: Or to clarify - we already do map msprime lineages to random slim nodes, but I guess removing the assertion allows for there to be more slim nodes than msprime nodes at the merger. We would have to set `replace=True` for `np.random.choice`if the msprime sample size were set larger than slim nodes alive at 0, which would map multiple msprime nodes to some slim nodes and would not guarantee full coverage of slim lineages in the merger.

petrelharp: Hm, right - this probably needs more explaining in the vignette. To make things concrete, let's say we run a SLiM simulation with a population size of 100 for a while, and then follow it up with 50 more generations of msprime simulation with 100 samples from an Ne of 100. So, the end of the SLiM simulation will be "50 generaitons ago", and what we're doing is we (a) start with 100 lineages, today; (b) run these back through time until 50 generations ago, and (c) match them up, randomly, with the individuals alive at the end of the SLiM simulation. In step (c), we match them *randomly* because under a neutral model, any of those individuals are as likely to be ancestors as any other ones. Note that the result of step (b) will be *fewer* than 100 lineages - probably quite a lot fewer at any particular locus - and so when we match them up, some SLiM individuals don't get matched to lineages. These are individuals that didn't leave behind any genetic material that made it until today (at the locus we're looking at). So, there's no problem having fewer lineages than SLiM individuals. Also note that there's no constraint in msprime regarding the number of samples and Ne - you could take a million samples from a population with Ne=100 if you wanted - this might represent juveniles from a population with very high juvenile mortality, for instance.

So, the only awkwardness I see is if we end up with *more* lineages at the end of the SLiM simulation than individuals. But I'm not sure we should disallow this, as maybe someone would have a good reason for doing things that way?

petrelharp:
Great! Well, I suppose the thing to do is to go open a PR for msprime. The PR should describe the use case and rationale (eg, by referring to the vignette), and mention any choices that you're uncertain about (if any) or questions. And, you can mark it as a 'draft' until you think it's ready. Are you comfortable doing that? I will have a look over there.

I'd recommend calling it `msprime.continue_simulation( )`, rather than a a method of the tree sequence. And, maybe there's a more descriptive name? As it is, it might be easy to mix up whether it "continues" towards the present or towards the past. Nothing better is coming to mind, though. I suppose one question will be whether to implement this in C or just keep it at the python layer; I think the latter is totally fine, so you should start there in any case, but if you're motivated to do it, we could move it into the C API. (There'd be no advantage to speed, it'd be only for completeness of the C API, really.)

It seems like you know what to do, but let me know if you'd like me to start this off!

pyslim/slim_tree_sequence.py
jelliotthiggins: That clarification helps a lot - seems like a non-issue overall, just requires some documentation. In the rare case of msprime having a higher count of nodes at the merger time, do you think it is better to a) prevent this case using an assertion or b) allow it by conditionally removing the replacement condition in `np.random.choice()`?

jelliotthiggins:
Perfect! I'll get the code moved over there and cleaned up and let you know when the PR is up. Thanks again!

jelliotthiggins:
@petrelharp I have a draft PR up on msprime now and I am working on getting good test coverage!

pyslim/slim_tree_sequence.py
petrelharp: Hm - thinking this through -`replace=False` is like saying that the msprime lineages *are the same individuals as* the final SLiM chromosomes; while `replace=True` is like saying there's a round of WF reproduction between the two. If the SLiM population size is much bigger than the number of lineages, the difference will be tiny. I like the `replace=False` interpretation better, but as you say, `replace=True` lets the number of lineages be greater than the population size (as you'd get from a sudded population size change in a WF model). 

So, I agree, changing this to `replace=True` is a good way to do it.

jelliotthiggins: I believe I found a better solution:

```
new_nodes = np.where(new_ts.tables.nodes.time == time)[0]

    if len(new_nodes) > len(old_nodes):
        new_nodes = np.random.choice(new_nodes, len(old_nodes), replace=False)
    node_map = np.repeat(tskit.NULL, new_ts.num_nodes)
    node_map[new_nodes] = np.random.choice(old_nodes, len(new_nodes), replace=False)
```

Basically, if the number of lineages is greater than the population size I drop extra lineages randomly and have an equal amount of lineages to the population size. How does that look to you? Either solution could work I think, this one just preserves a one-to-one ratio of lineages between the tree sequences.

petrelharp: Hm, this is a bit tricky. It doesn't work to drop lineages, since that would imply that some of the samples in the resulting tree sequence don't inherit from *anyone* past this time.  Also, `node_mapping` is - I think  - required to be one-to-one, so that we can't have two lineages mapipng to a single individual (but [this needs figuring out](https://github.com/tskit-dev/tskit/blob/c9b2c920b6c4e1aaebcbc91dc4c6e13e919bdd08/c/tskit/tables.c#L9400) - and if so, `replace=True` isn't allowed.

jelliotthiggins: Ah that makes sense. I think the solution for now may have to be to disallow it then. Having `replace=True` sets a not great precedent and could cause issues down the road, and yes having samples that don't inherit from anyone past that time is wrong.

petrelharp: It seems unlikely to be a restriction: if the number of samples is less than or equal to the size of the final SLiM population, then we'd never have too many samples, anyhow.

PR #125, opened 2021-01-17
Support retain as well as remember

hyanwong:
Hi @petrelharp - here's the basic PR for dealing with Retaining in pyslim *but* I haven't added any more tests. I guess we should add a slim recipe in where e.g. we Retain (rather than remember) everyone, and maybe a recipe where we Retain some and Remember others? I guess we want a round trip of some sort - does this exist already?

Also, we are still testing the (deprecated) function `first_generation_individuals()`, but that function relies on the `INDIVIDUAL_FIRST_GEN` constant that no longer exists. What should we do this this: simply turn the warning into an error? It's causing `tests/test_tree_sequence.py::TestDeprecations::test_first_gen` to fail.


docs/_static/Makefile
```diff
@@ -34,7 +34,7 @@ pedigree3.svg : pedigrees.ink.svg
 pedigree_recapitate.svg : pedigrees.ink.svg
 	./export-layers-svg.sh $< \
 		ts_indivs simp_indivs pedigree simp_pedigree chromosomes simplify_removes \
-		recap_pedigree extra_chromosomes tskit_time recap_label recap \
+		recap_pedigree tskit_time recap_label recap \
```
petrelharp, line 37: Did you mean to do this? Here's the change:
Screenshot from 2021-01-18 08-37-21
Screenshot from 2021-01-18 08-37-28

docs/introduction.rst
```diff
@@ -65,23 +67,32 @@ in which many of the nodes and individuals have been removed.
 Who and what is in the tree sequence?
 *************************************
 
-OK, who and what exactly *is* left in the tree sequence?
+OK, who and what exactly *is* left in the tree sequence after simplification?
```
petrelharp, line 70: It migth be confusing to say "after simplification" here: the reader might not know that "simplification" refers to a specific process (the previous paragraph is understandable without thinking of it as a technical term).

petrelharp: Maybe "who and what exactly *is* left in the tree sequence after the unnecessary information has been removed?" (or, "has been simplified away"?)

docs/introduction.rst
```diff
@@ -65,23 +67,32 @@ in which many of the nodes and individuals have been removed.
...
 We clearly want all individuals of the *final* generation
 to have their genealogies recorded in the tree sequence.
-in the picture above there is one additional individuals pictured
```
petrelharp, line 71: whoops, should have capitalized "in"?

petrelharp:
> Also, we are still testing the (deprecated) function first_generation_individuals(), but that function relies on the INDIVIDUAL_FIRST_GEN constant that no longer exists. What should we do this this: simply turn the warning into an error? It's causing tests/test_tree_sequence.py::TestDeprecations::test_first_gen to fail.

I'd say we should keep `INDIVIDUAL_FIRST_GEN` but set it equal to `INDIVIDUAL_RETAINED`, with a comment that it's deprecated, but kept for backwards compatibility?

docs/introduction.rst
```diff
@@ -65,23 +67,32 @@ in which many of the nodes and individuals have been removed.
...
+tree sequence that were not alive at the end of the simulation, because they
+were *remembered*, using the ``treeSeqRememberIndividuals()`` method. The lower
+one (still containing 2 nodes or chromosome) was *permanently* remembered, which is the
+default behaviour. The upper one (with only one node left) was simply marked as *retained*,
+done by passing ``permanent=F`` when calling ``treeSeqRememberIndividuals()``.
+As you might expect, permanently remembered individuals and their nodes are always
+kept in the tree sequence after simplification. In contrast, individuals that are
+merely marked for retention can lose their nodes (as has happened to one of the nodes in
+this case). Furthermore, once all an individuals's nodes are lost, simplification
+removes the individual too, for efficiency reasons.
```
petrelharp, line 83: Suggested edit:
"
In the picture above there are two additional individuals pictured in the
tree sequence that were not alive at the end of the simulation. These are still present because they
were *remembered*, using the ``treeSeqRememberIndividuals()`` method. The lower
one (still containing 2 nodes or chromosome) was *permanently* remembered, which is the
default behaviour. The upper one (with only one node left) was only "retained",
by passing ``permanent=F`` when calling ``treeSeqRememberIndividuals()``.
As you might expect, permanently remembered individuals and the entire genealogical history of both their nodes are always
kept in the tree sequence, while individuals that are
merely retained will only be present in the tree sequence as long as at least one of their nodes
has not been removed by simplification.
"

Hm. I'm kinda feeling like this is the wrong place for a technical discussion, being the *introduction*, and maybe "retaining" is a more advanced topic to be explained elsewhere. What do you think? The problem is that the above paragraph still isn't quite there, because the fundamental difference between permanent/not is that 
- a permanently remembered individual is guaranteed to always have the entire history of both their chromosomes in the tree sequence
- whereas a retained individual only acts as some extra annotation for its nodes: the only parts of its genomes that are actually recorded in the tree sequence are those on which it's the MRCA of two samples, unless `keep_unary` is on.

How about we remove mention of "retaining" from the Introduction, and make a new section in the Tutorial about it?

docs/tutorial.rst
```diff
@@ -939,10 +943,17 @@ Also known as "gotchas".
 
 2. Make sure to distinguish *individuals* and *nodes*!
    ``tskit`` "nodes" correspond to SLiM "genomes".
-   Individuals in SLiM are diploid, so each has two nodes.
+   Individuals in SLiM are diploid, so normally, each has two nodes (but see below for
+   retained individuals, which can lose one of the nodes during simplification).
```
petrelharp, line 947: how about 
```
   Individuals in SLiM are diploid, so each has two nodes (but retained individual may have
   some removed by simplification: see below).
```

pyslim/slim_tree_sequence.py
```diff
@@ -627,7 +627,7 @@ def individuals_alive_at(self, time, stage='late', remembered_stage=None,
         :param str stage: The stage in the SLiM life cycle that we are inquiring
             about (either "early" or "late"; defaults to "late").
         :param str remembered_stage: The stage in the SLiM life cycle
-            that individuals were Remembered during (defaults to the stage the
+            that individuals were Remembered or Retained (defaults to the stage the
```
petrelharp, line 630: maybe omit "or Retained" here? Since one retains by calling treeSeqRemember, it seems sufficient; and otherwise it suggests there is a treeSeqRetain method?

pyslim/slim_tree_sequence.py
```diff
@@ -714,8 +714,8 @@ def individual_ages_at(self, time, stage="late", remembered_stage="late"):
         :param float time: The reference time ago.
         :param str stage: The stage in the SLiM life cycle used to determine who
             is alive (either "early" or "late"; defaults to "late").
-        :param str remembered_stage: The stage in the SLiM life cycle that
-            individuals were Remembered during.
+        :param str remembered_stage: The stage in the SLiM life cycle during which
+            individuals were Remembered or Retained.
```
petrelharp, line 718: ditto

petrelharp:
> I guess we should add a slim recipe in where e.g. we Retain (rather than remember) everyone, and maybe a recipe where we Retain some and Remember others? 

Yep - just adding the recipe, and then adding it to [example_files](https://github.com/tskit-dev/pyslim/blob/main/tests/__init__.py#L28) should do it.

docs/_static/Makefile
hyanwong: I did mean to do this because I thought you would want the picture in the recapitated sequence to match the ones in part B of pedigree01.png? It seemed very odd that it included the extra (simplified-away) chromosomes. Perhaps I misunderstood something?

docs/introduction.rst
hyanwong: Good point. On that note, when you mention simplification later on, in `tutorial.rst`, it's as if you have never mentioned it before. I wonder if, in the tutorial, you should say something like "explicit simplification, in a similar way that the tree sequence is simplified on output, as described in the intro"?

docs/_static/Makefile
petrelharp: Ah, I see. Good call.

docs/introduction.rst
petrelharp: Good idea, this should be clarified. In the tutorial I'm talking about running-simplify-in-python, not the simplification that happens in SLiM as it goes along. The reason it's like this is because simplification-by-SLiM is something that I think people can safely ignore: it's how things work under the hood, but I find that if we start talking about simplification a lot then people think *they* have to do it, but things are less error-prone if they don't do it. But clarification would be good.

docs/introduction.rst
hyanwong: I think it's useful to have it for the picture (because we need a picture for showing the difference between retained and remembered).

Also, if we are removing mention of retaining, should we not also remove a mention of Remembering entirely from the intro, as many people will use pyslim without Remembering or Retaining?

I wonder if we should make Remembering (+ retaining) a separate section on their own, perhaps in the intro?

petrelharp:
Want me to have a go at the tests here, while you finish editing the docs?

hyanwong:
> Want me to have a go at the tests here, while you finish editing the docs?

If that's easy, yes please!

If we are adding `KEEP_UNARY_IN_INDIVIDUALS` too, then you might want to bear that in mind when writing the tests, to make it easier to add tests for that too?

hyanwong:
@petrelharp - I've had a go at rewording `introduction.rst` as we discussed, although I haven't yet changed the diagrams to match: I hope you can imagine what they might look like.

The section "What if I'm interested in historical individuals too?" is in the intro, but could, if you prefer, be moved to the tutorial. I'll wait until I've got the go ahead from you before making any more changes or fiddling with the plots, if that's OK?

docs/introduction.rst
```diff
@@ -11,106 +11,143 @@ or modifying files produced by other programs (e.g.,
 `msprime <https://msprime.readthedocs.io/en/stable/>`_,
 `fwdpy11
 <https://fwdpy11.readthedocs.io/en/stable/pages/tsoverview.html>`_
-and `tsinfer <https://tsinfer.readthedocs.io/>`_) for use in SLiM. 
+and `tsinfer <   8555https://tsinfer.readthedocs.io/>`_) for use in SLiM.
```
petrelharp, line 14: I assume this was on accident?

docs/introduction.rst
```diff
@@ -11,106 +11,143 @@ or modifying files produced by other programs (e.g.,
...
-To do this, SLiM has added several additional types of information to the basic tree sequence file.
+To do this, SLiM has added several additional types of information to the basic
+tree sequence file.
 
 ********
 Overview
 ********
 
-A tree sequence is a way of storing both genealogies and genotypes
+A tree sequence is a way of storing both the full genetic history and the genotypes
```
petrelharp, line 28: nice

docs/introduction.rst
```diff
@@ -11,106 +11,143 @@ or modifying files produced by other programs (e.g.,
...
-merely marked for retention can lose their nodes (as has happened to one of the nodes in
-this case). Furthermore, once all an individuals's nodes are lost, simplification
-removes the individual too, for efficiency reasons.
-
+OK, who and what exactly *is* left in the tree sequence after the unnecessary
+information has been removed? Here's the diagram above with some added terminology. Note
+that for simplicity, in this model the individuals that are currently alive are
+all in the final generation, but more generally, individuals alive at a particular time
+should be identified via the :meth:`.SlimTreeSequence.individuals_alive_at()` method (see
+:ref:`sec_extracting_individuals`).
```
petrelharp, line 77: this note is getting in to the weeds a bit much - in the diagram it isn't exactly clear who's alive when, or if this is something with overlapping generations or not

docs/introduction.rst
```diff
@@ -11,106 +11,143 @@ or modifying files produced by other programs (e.g.,
...
+mark additional nodes as samples: see :ref:`sec_remembering_individuals`)
+
+In contrast to the individuals containing sample nodes, you can see that all the other
+circles, representing historical (i.e. dead) individuals, have gone, although sometimes
+their nodes remain. By default, only individuals with sample nodes are recorded in the
+tree sequence; that means ancestral nodes lose any information about which individuals
+they were in (see :ref:`sec_retaining_individuals` if you need to retain this information).
+
+As well as the actual individuals, many historical *nodes*  have been removed too,
+along with the lines linking them to the rest of the genealogy (in tree-sequence-speak the
```
petrelharp, line 95: maybe "along with their genealogical relationships"

docs/introduction.rst
```diff
@@ -11,106 +11,143 @@ or modifying files produced by other programs (e.g.,
...
+lines are known as "edges"). The deleted nodes are simply those that are not needed to
+reconstruct the relationships between the sample nodes. For example we remove nodes
+leading to a dead end (e.g. in individuals who had no offspring). Similarly, as time goes
+on, successive recombination events can gradually reduce the genetic contribution of
```
petrelharp, line 99: it's genetic drift rather than recombination events that does this, strictly speaking

docs/introduction.rst
```diff
@@ -11,106 +11,143 @@ or modifying files produced by other programs (e.g.,
...
+``retainCoalescentOnly`` parameter of ``treeSequenceInitialize()``)
+
+*****************************************************
+What if I'm interested in historical individuals too?
+*****************************************************
+
+As we've seen, the tree sequence retains enough of the nodes (genomes) in the past to
+reconstruct the genetic ancestry of the currently alive individuals. But you might want
+more than that. For example, there may be individuals who are not alive any more, but
+whose complete ancestry you would like to know.  Or it may be that you want to access the
```
petrelharp, line 115: ", or you'd like to know how the final generation relates to the individuals alive at some previous time."

docs/introduction.rst
```diff
@@ -11,106 +11,143 @@ or modifying files produced by other programs (e.g.,
...
+spatial location of genomes in the past (which, for historical reasons is linked to
+individuals, not to genomes). This is done by *remembering* an individual during the
+simulation, using the SLiM function ``treeSeqRememberIndividuals()``. 
+
+.. _sec_remembering_individuals:
+
+By default, individuals are *permanently* remembered, by marking their nodes as actual
```
petrelharp, line 122: should be "individuals included in the tree sequence by a call to ``treeSeqRememberIndividuals( )``"

docs/introduction.rst
```diff
@@ -11,106 +11,143 @@ or modifying files produced by other programs (e.g.,
...
 .. figure:: _static/pedigree3.png
    :scale: 40%
 
-.. note::
-
-   In previous versions of SLiM/pyslim, the *first generation* of individuals
-   were kept in the tree sequence, to allow recapitation.
-   This is no longer necessary, so these are *no longer present*, unless
-   you specifically Remember them -- see the :ref:`sec_tutorial_recapitation`
-   section of the tutorial for more discussion of how this works.
```
petrelharp, line 101: I think we could leave this note in for the time being?

petrelharp:
I've made some minor comments on the docs - they look good, generally. I'd still vote to move the "historical individuals" bit to a new section in the Tutorial, maybe before "Extracting particular individuals". What do you think? I'll have another good read through after the figures are in and it compiles.

docs/introduction.rst
hyanwong: Oops! I was fixing a USB keyboard at the same time and didn't realise it was still plugged in to my laptop. 😆

docs/introduction.rst
hyanwong: OK. It was in the previous write-up, but I wasn't quite sure where to slide it in. Note that the diagram isn't right yet - I haven't changed anything there.

docs/introduction.rst
hyanwong: Oh yes, I forgot about putting that back in, sorry. I'll find a place for it. If we move the "historical individuals" bit to the tutorial (which I'm not averse to), then perhaps that's a good place to put it, so it's not so in-your-face in the introduction.

petrelharp:
Ok, I've added some tests, and fixed up the testing of `individual_parents` (a bit tricky, logicking through when exactly we should know about someone's parents or not).

And, this has turned up - I think - a bug in the SLiM implementation: the test recipe where [I alternate retaining and remembering](https://github.com/tskit-dev/pyslim/pull/125/commits/f693cdef3d774fa7840fbf8bc2e3b9d776b9b38c#diff-5282d3706043177f54afced9c850b8a422af30cbc6c2ac4a0d9f1385afc63790R33), `tests/examples/recipe_remember_and_retain.slim` produces a tree sequence in which (at least) one individual, individual 21 in the tree sequence, with SLiM ID 28, is Remembered, but has *only one node*, which AFAICT, should not happen, since Remembered individuals should always have both nodes. This individual was both Remembered and Retained, so I'm wondering if perhaps we've forgotten to re-mark individuals as samples appropriately when they are Retained after Remembering? Any chance you could see what's going on here, @hyanwong?

docs/introduction.rst
petrelharp: figured, but wondered if it was some advanced URL magic I didn't know about =)

hyanwong:
> And, this has turned up - I think - a bug in the SLiM implementation: the test recipe where [I alternate retaining and remembering](https://github.com/tskit-dev/pyslim/pull/125/commits/f693cdef3d774fa7840fbf8bc2e3b9d776b9b38c#diff-5282d3706043177f54afced9c850b8a422af30cbc6c2ac4a0d9f1385afc63790R33), `tests/examples/recipe_remember_and_retain.slim` produces a tree sequence in which (at least) one individual, individual 21 in the tree sequence, with SLiM ID 28, is Remembered, but has _only one node_, which AFAICT, should not happen, since Remembered individuals should always have both nodes. 

Yes, I agree.

> This individual was both Remembered and Retained, so I'm wondering if perhaps we've forgotten to re-mark individuals as samples appropriately when they are Retained after Remembering? Any chance you could see what's going on here, @hyanwong?

Yes, I'll check this out.

docs/introduction.rst
hyanwong: ISWYM, but here I think it's a combination of both. What I'm trying to convey is that even if there are lines of ancestry in the diagram that go through a (coalescent) node, that node may be deleted because the line essentially "belongs to another tree". I'l try to rephrase.

hyanwong:
> > This individual was both Remembered and Retained, so I'm wondering if perhaps we've forgotten to re-mark individuals as samples appropriately when they are Retained after Remembering? Any chance you could see what's going on here, @hyanwong?
> 
> Yes, I'll check this out.

I've just rebased on the new test framework, and the tests are all passing, so I can't find the problem you say, @petrelharp. Could you check if I'm missing something? It should be easier to try out a failing test & recipe now, so hopefully easier to report?

petrelharp:
Uh-oh, I'm afraid something's gone wrong with your rebase? For some reason this is showing a bunch of fixture-related changes in the [diff](https://github.com/tskit-dev/pyslim/pull/125/files).  And (maybe this was an error that happened in that PR that we missed) there's a bunch of tests in `TestIndividualAges` that are duplicated, from `test_population` on down.

petrelharp:
The fixture commits showing up here is I think because you rebased to your local feature branch instead of rebaseing to `origin/main`, which has a merge commit: 4fa846bea0c15e9c06cec943671fdf5291636e29 (for some reason).

petrelharp:
But it looks like the merging/rebasing might need to be a bit more manual. I've got the previous state of this branch in my git repo if you want me to force-push it back to how it was before the rebase.

hyanwong:
Oh, damn. Sorry. I did spot some problems and am trying to fix them in the SLiM code, but it seems like it's a bit fragile, so I'm trying to get https://github.com/MesserLab/SLiM/pull/142 fixed first, using the same approach in the pyslim tests. Might take a day or so to remember all the logic, though! Then I'll get round to this one.

hyanwong:
> But it looks like the merging/rebasing might need to be a bit more manual. I've got the previous state of this branch in my git repo if you want me to force-push it back to how it was before the rebase.

Erm, maybe? I don't quite understand this stuff though. I'm not sure how I go from here. Do I get you to rebase your stuff against the new `origin/main`, make a copy of my current branch, then rebase it against your branch (that I pull onto my local repo) perhaps? I'm a bit confused as to how to go from here.

petrelharp:
I could force-push to this branch, putting it back to the state it was in before you tried to rebase, and you could have a do-over by resetting your local branch. If it were me, I'd want to have a do-over, and wasn't sure if you had the original state before rebasing still handy.

petrelharp:
I've put it up here: https://github.com/petrelharp/pyslim/tree/old-support-retain in case it's useful.

hyanwong:
> I've put it up here: https://github.com/petrelharp/pyslim/tree/old-support-retain in case it's useful.

Perfect, thanks. I've hopefully just re-merged it into this PR, and squashed *everything* together, as it was getting to be such a mess. I now get the failure:

```
tests/test_tree_sequence.py:564: AssertionError
----------------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------------
Individual does not have two nodes, who should:
{'id': 21, 'flags': 393216, 'location': array([0., 0., 0.]), 'nodes': array([69], dtype=int32), 'population': 1, 'time': 7.0, 'metadata': {'pedigree_id': 28, 'age': 1, 'subpopulation': 1, 'sex': -1, 'flags': 0}}
{'parents': (8, 2), 'age': {(2, 'early'): 0, (2, 'late'): 0, (3, 'early'): 1}}
=========================================================================== short test summary info ============================================================================
FAILED tests/test_tree_sequence.py::TestHasIndividualParents::test_pedigree_parents[recipe_remember_and_retain.slim] - AssertionError: assert 1 == 2
```

So I'll look into it.

hyanwong:
NB: fixes #134

hyanwong:
OK, so I think that https://github.com/MesserLab/SLiM/pull/150 fixes the bug you noticed @petrelharp (my bad), and I've added another check into your tests with the latest commit. But when running `pytest tests/test_tree_sequence.py::TestHasIndividualParents::test_pedigree_parents -k recipe_remember_and_retain.slim`, I'm now getting the flooring. Any idea what's going on?

```
            if len(slim_p) != 2:
                # if we don't have two slim parents, then we sure shouldn't
                # have any pyslim parents
>               assert ts_p == []
E               assert [52] == []
E                 Left contains one more item: 52
E                 Use -v to get the full diff

tests/test_tree_sequence.py:581: AssertionError
=========================================================================== short test summary info ============================================================================
FAILED tests/test_tree_sequence.py::TestHasIndividualParents::test_pedigree_parents[recipe_remember_and_retain.slim] - assert [52] == []

```

petrelharp:
Hm, I don't know - options are (a) I haven't logicked through the conditions in the test exactly right, or (b) another bug. In that line there, `slim_p` are the SLiM IDs of the parents as output in the .pedigree file; there should be two entries in it for anyone produced by SLiM (even if it was by selfing; we'd have the same parent twice then). And the `individual_parents` method only gives parents if the entire diploid genome is accounted for in one or two parent individuals, which should not happen if it doesn't have SLiM parents.

docs/installation.rst
```diff
@@ -4,7 +4,7 @@
 Installation
 ============
 
-Tszip can be installed using the standard `pip <https://pypi.org/project/pip/>`_ distribution method::
+Pyslim can be installed using the standard `pip <https://pypi.org/project/pip/>`_ distribution method::
```
petrelharp, line 7: nice catch here, lol.

hyanwong: At least now that fact that it's nicked from tszip is a bit less obvious

> But plagiarize, plagiarize, plagiarize
> Only be sure always to call it please 'research'

https://www.youtube.com/watch?v=gXlfXirQF3A

:)

petrelharp:
Something happened here, btw: running the tests on this branch I get
```
FAILED tests/test_provenance.py::TestProvenance::test_convert_0_1_files - FileNotFoundError: [Errno 2] No such file or directory: '/home/peter/projects/tskit-dev/pyslim/tests/test_recipes/recipe_WF.v3.0.trees'
FAILED tests/test_provenance.py::TestProvenance::test_convert_0_2_files - FileNotFoundError: [Errno 2] No such file or directory: '/home/peter/projects/tskit-dev/pyslim/tests/test_recipes/recipe_WF.v3.2.trees'
FAILED tests/test_provenance.py::TestProvenance::test_convert_0_3_files - FileNotFoundError: [Errno 2] No such file or directory: '/home/peter/projects/tskit-dev/pyslim/tests/test_recipes/recipe_WF.v3.3.1.trees'
FAILED tests/test_provenance.py::TestProvenance::test_convert_0_4_files - FileNotFoundError: [Errno 2] No such file or directory: '/home/peter/projects/tskit-dev/pyslim/tests/test_recipes/recipe_WF.v3.4.trees'
```

hyanwong:
Oh, weird. Are those files missing from your version then? They seem to be in mine.

hyanwong:
Oh, I see: `.trees` files are excluded in the .gitignore, so I didn't notice. Will fix.

hyanwong:
Done: should work now. Sorry about that.

petrelharp:
Maybe we should mark those particular .trees files to *not* be ignored in the gitignore...

petrelharp:
Ok, so I get the same error. (tldr; we should delete this part of the test)

Adding some prints and running the SLiM script and looking at the pedigree file, I see that this individual's ID is 77, and has two SLiM parents, 52 and 64. 
```
generation      stage   individual      age     parent1 parent2
...
8       early   77      0       64      52
```
But for some reason 64 isn't in `slim_p`. With some more prints, it turns out that 77 happened to inherit all of *both* its genomes from 52, and none from 64. Huh, that shouldn't happen, under diploidy??
```
individual: {'id': 0, 'flags': 458752, 'location': array([0.22473303, 0.98065234, 0.        ]), 'nodes': array([70, 71], dtype=int32), 'population': 1, 'time': 2.0, 'metadata': {'pedigree_id': 77, 'age': 2, 'subpopulation': 1, 'sex': -1, 'flags': 0}}
parent: {'id': 36, 'flags': 393216, 'location': array([0., 0., 0.]), 'nodes': array([50, 51], dtype=int32), 'population': 1, 'time': 5.0, 'metadata': {'pedigree_id': 52, 'age': 2, 'subpopulation': 1, 'sex': -1, 'flags': 0}}

{left=0.000, right=10.000, parent=50, child=70, id=69, metadata=b''}
{left=32.000, right=55.000, parent=50, child=70, id=70, metadata=b''}
{left=70.000, right=100.000, parent=50, child=70, id=71, metadata=b''}
{left=10.000, right=32.000, parent=51, child=70, id=75, metadata=b''}
{left=55.000, right=70.000, parent=51, child=70, id=76, metadata=b''}
{left=0.000, right=53.000, parent=50, child=71, id=72, metadata=b''}
{left=53.000, right=100.000, parent=51, child=71, id=77, metadata=b''}

```
Ah-ha. **but** that other parent, with SLiM ID 64, is itself produced by 77's other parent, 52, by selfing:
```
generation      stage   individual      age     parent1 parent2
7       early   64      0       52      52
```
and I guess that 64 was just retained, not remembered, and so isn't in the tree sequence any more. The result is indistinguishable from 77 having been produced by 52 by selfing.

Ok: my proposed fix is to just *delete* the lines
```
            if len(slim_p) != 2:
                # if we don't have two slim parents, then we sure shouldn't
                # have any pyslim parents
                assert ts_p == []
```
since they're trying to test something we can't really test.

hyanwong:
> Maybe we should mark those particular .trees files to _not_ be ignored in the gitignore...

Nah, it's OK. Once they are in the git store, it doesn't ignore them - if they change it will now be noticed. They only vanished because I changed the name of the directory they were in, so git forgot they were actually part of the repo.

hyanwong:
> my proposed fix is to just _delete_ the lines ... since they're trying to test something we can't really test.

Great sleuthing. Thanks @petrelharp. Happy with that solution. I'll do it now.

hyanwong:
Interesting that this is now failing on

```
python -m pytest tests/test_tree_sequence.py::TestHasIndividualParents -k recipe_remember_and_retain.slim
```

This is even with the updated SLiM version (and we aren't doing keep_unary in the tests anyway). Any idea what's going on @petrelharp ? I'm just working on the Inkscape pictures for the new docs now.

hyanwong:
I have now changed the docs so that "remembering and retaining" has its own section. That's meant a fair change in the diagrams, and a bit in the structure (e.g. I now have put "recapitation, simplification, mutation" in a higher-level section about combining forwards & backwards sims.

I rather hacked the diagrams so haven't properly changed the master Inkscape file or the makefile that creates them. I'll do that once you look at the pictures, @petrelharp, and approve the general approach. I hope it makes the introduction a bit simpler (I wasn't sure which style out of `pedigree2.png` or `pedigree3.png` to retain, though: solid circles with coloured labels, or labels with arrows. It did seem like having 2 essentially identical diagrams wasn't needed any more, though)

petrelharp:
Well, the failing test is the same problem as above: individual 78 has parents 62 and 65, but only 62 is retained, and so we mistake 65's parent, 52, for 78's parent, because apparently 52 is still alive when 78 is born, and 78 inherited the entire chromosome from 52. So, overly strict tests again, I guess - I only really designed these `individual_parents( )` methods for when everyone is remembered.

Here's a fix:
```
diff --git a/tests/test_tree_sequence.py b/tests/test_tree_sequence.py
index af41e07..c81e8e0 100644
--- a/tests/test_tree_sequence.py
+++ b/tests/test_tree_sequence.py
@@ -576,9 +576,13 @@ class TestHasIndividualParents(tests.PyslimTestCase):
                 # should have parents when individual and both parents are remembered or alive
                 for b in definitely_slim_p:
                     assert b in ts_p
-            for a in ts_p:
-                # all pyslim parents should be legit
-                assert a in slim_p
+            # all pyslim parents should be legit (so set(ts_p) - set(slim_p) should usually be empty)
+            # BUT sometimes we can mistake a grandparent for a parent
+            gfolks = []
+            for a in set(info[sid]["parents"]) - set(ts_p):
+                gfolks.extend(info[a]["parents"])
+            for a in set(ts_p) - set(slim_p):
+                assert a in gfolks
 ```

File: docs/introduction.rst [hunk 17]
```diff
@@ -13,92 +13,96 @@ or modifying files produced by other programs (e.g.,
...
+   :align: right
+
+In the recorded tree sequence the alive individuals, which in this model are those
+in the final generation, have all had their nodes marked as *samples*. As
+you may recall, this means we have their full genetic ancestry. The sample nodes,
+and the individuals containing them, are always present in the tree sequence, which
+should be clear in the diagram. (Note that additional nodes can be
+:ref:`marked as samples <sec_remembering_individuals>` as detailed in the tutorial.)
```
petrelharp, line 84: I'd end this paragraph with "The sample nodes, and the individuals containing them, are always present in the tree sequence." (the rest is a bit confusing at this point).

docs/introduction.rst
```diff
@@ -13,92 +13,96 @@ or modifying files produced by other programs (e.g.,
...
-We clearly want all individuals of the *final* generation
-to have their genealogies recorded in the tree sequence.
-in the picture above there is one additional individuals pictured
-in the tree sequence that was not alive at the end of the simulation,
-because it was *remembered*, using the ``treeSeqRememberIndividuals()`` method.
-
+OK, who and what exactly *is* left in the tree sequence after the unnecessary
+information has been removed? Here's the diagram above with some added terminology.
 
 .. figure:: _static/pedigree2.png
```
petrelharp, line 75: Hm, there's "retained" and "remembered" individuals in the diagram, but not mentioned in the text. Oh:it says "all ... dead individuals have vanished", contradicting this - maybe you meant to put a different diagram here?

petrelharp:
Gee, you're right about those two figures having different color schemes. I think having the "reminder of the terminology" figure is useful, though - mostly to drive home "node" versus "individual" (a common source of errors!!) and also maybe "sample node". And, having it stand a bit apart is I think OK - it's a reminder, for reference. (pasting that figure here for context)
Screenshot from 2021-02-04 20-48-00

docs/tutorial.rst
```diff
@@ -203,10 +206,11 @@ and making sure that no recombinations occurred in that region.
    the RecombinationMap class; setting ``discrete=True`` will more closely
    match the recombination model of SLiM.
 
+.. _sec_tutorial_simplification:
 
-**************
+++++++++++++++
 Simplification
```
petrelharp, line 212: This and the next section ("adding mutations") aren't really part of "combining forward and coalescent simulations" - they're things you'd want to do even without any coalescent simulations at all. I don't think we need the "Combining..." top-level section at all?

docs/tutorial.rst
```diff
@@ -320,11 +328,102 @@ functions that return tree sequences may be converted back to
...
+
+
+.. _sec_remembering_individuals:
+
+Permanently remembering individuals
+-----------------------------------
+
+By default, a call to ``treeSeqRememberIndividuals()`` will permanently remember one or
+more individuals, by marking their nodes as actual samples: the simulated equivalent of
+ancient DNA dug out of permafrost, or stored
```
petrelharp, line 364: nice

petrelharp:
Whoops - I was looking at an out-of-date version, and have deleted a few comments above. Looking good! I like the new Introduction figure!

petrelharp:
> although it's on the "tutorial" page, the "remembering and retaining" section doesn't actually have any code to follow. Which seems a bit odd.

Good point. And the subsequent sections ("Extracting particular individuals" and "Individual metadata") should arguably go earlier, because they describe more basic things people would want to do. Maybe we should put in a simple example that uses remembering or retaining? e.g. just Remember everyone halfway through a nonWF simulation and then report the age structure at the time. I could throw tihs in a separate PR if you like.

petrelharp:
Looks good! I approve!

hyanwong:
> Well, the failing test is the same problem as above: individual 78 has parents 62 and 65, but only 62 is retained, and so we mistake 65's parent, 52, for 78's parent, because apparently 52 is still alive when 78 is born, and 78 inherited the entire chromosome from 52. So, overly strict tests again, I guess - I only really designed these `individual_parents( )` methods for when everyone is remembered.

Ah, embarrassed to say that I didn't follow too deeply what the problem was, so I didn't spot this, sorry (I was focussing on the docs instead) Thanks for the fix. It all passes now. I'll address your revised comments above, fix the SVG, and hopefully this will be good to go, so I can move on to the final tskit fix.

docs/introduction.rst
hyanwong: Good point, I was probably going too deep too early.

docs/tutorial.rst
hyanwong: Ah, that was because the entire 3-part section has its own introduction. Previously that was the introduction used for the entire tutorial, but it didn't really fit because it was only relevant to the "recapitation -> simplification -> mutation" part. I guess a simple way to change this is to change the section heading to "recapitation, simplification, and mutation". But perhaps there's a better way?

petrelharp: Right. This is a symptom of the docs being just cobbled together, and needing an overhaul. Those three things are together because they're part of the expected workflow. Maybe a heading titled "The standard workflow"? Or, "Common preprocessing steps"? Or, "The first things to do with your tree sequence"? Or, maybe just "Recapitation, simplification, and mutation", as you say; we'll rearrange later.

hyanwong: I changed it to "Recapitation, simplification, and mutation", and added a bit of intro to explain that this is a common workflow.

I have overhauled some of the other sections a bit, and I think that makes them a tiny bit less of a hotchpotch. But I don't want to do much more (especially if we are going to move this over to a jupyter-book anyway).

hyanwong:
This all looks *great* @petrelharp thanks for all your work on this and the associated fixes/tests in https://github.com/MesserLab/SLiM/pull/158 and https://github.com/tskit-dev/tskit/pull/1190 . I didn't realise how much work it would be to get all these moving parts working together 🤯 But it looks like it's all done now, so 🎉

petrelharp:
Whoo! I think it's working now! Tests passed locally. I think I'll be ready to squash & merge. Let me know if you spot anything.

hyanwong:
LGTM. Squash & merge away! (maybe even merge https://github.com/tskit-dev/tskit/pull/1190 in tskit beforehand?)

PR #131, opened 2021-01-24
Refactor tests

hyanwong:
This is only a draft of what can be done, so let me know if it looks right @petrelharp, and I can try rolling it out to the other tests too. Only the `test_tree_sequence.py::TestSlimTime` test has been properly altered to use the `basic_slim_recipe` fixture, but it does work entirely in parallel, both over the tests and over the recipes, e.g. if you do

```
pytest test_tree_sequence.py::TestSlimTime -n 4
```

for all the recipes, or just 

```
pytest test_tree_sequence.py::TestSlimTime -k nonWF_recipe
```
for one of them. Annoyingly the way we decided to do it means that we call the test function for every script, but simply do nothing if the script doesn't need this test running on it. I suspect that means the total count of tests run will be over inflated . Perhaps there's a pytest way to decide, programmatically within the test, to say "skip this test and don't count it in the totals" (https://github.com/pytest-dev/pytest/issues/3730 seems relevant)?

The general setup is mostly now done in conftest.py, and similar to what I did in SLiM. the `slim_outfiles()` function defines the file names, SLiM, variable names, and a function to apply if the file is produced. Instead of the tree files having the name of the recipe, by default they are all called `out.trees`, but they are *stored* in a directory named by the recipe instead. I pass 2 SLiM variable names in, one for the trees file and another for the pedigree, but the second isn't always used by the script (which is fine, I would have thought).

There's a FIXME because I'm not sure where in the code we should normally set the random seed if these tests are all parallelized.

The only fixture so far is `basic_slim_recipe`. I intend to have another one for `restarted_slim_recipe`, but I haven't yet worked out how I'll do the restarting thing, as it will rely on the previous recipe having been run.

Also, I have had to change the names of all the recipes to put "recipe" as a suffix, rather than a prefix (i.e. `nonWF_recipe` rather than, as previously, `recipe_nonWF`), because the `-k` flag is always a partial matching thing, so that if you do `-k recipe_nonWF` it will run both `recipe_nonWF` and `recipe_nonWF_early`. I hope the name change is OK.


hyanwong:
> Perhaps there's a pytest way to decide, programmatically within the test, to say "skip this test and don't count it in the totals" ([pytest-dev/pytest#3730](https://github.com/pytest-dev/pytest/issues/3730) seems relevant)?

Actually, I've just found a *much* nicer way to do all this, using `@pytest.mark.parametrize` to restrict which recipes we run the tests on:

```
class TestHasIndividualParents(tests.PyslimTestCase):

    pedigree_recipes = [k for k, v in basic_slim_recipe_specs.items() if "pedigree" in v]

    @pytest.mark.parametrize('basic_slim_recipe', pedigree_recipes, indirect=True)
    def test_pedigree_has_parents(self, basic_slim_recipe):
         # this will be called only for recipes listed with {'pedigree': True}
         ts = basic_slim_recipe["ts"]
         # ts is a ts with an associated pedigree file
```

recipe_nonWF_early.trees.pedigree
```diff
@@ -0,0 +1,229 @@
+generation	stage	individual	age	parent1	parent2
```
petrelharp, line 1: let's not commit these temporary files produced by tests!

tests/README.md
```diff
@@ -0,0 +1,25 @@
+Install the dependencies using
+
+`pip install -r requirements/CI/requirements.txt`
```
petrelharp, line 3: if you're saying `python3` below don't you need to also say `pip3`?

recipe_nonWF_early.trees.pedigree
hyanwong: Oops. Well spotted. Must have been left from a previous trial. Sorry!

tests/README.md
hyanwong: Ah, possibly. No harm in that. Will change.

tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+```
+python3 -m pytest tests/test_tree_sequence.py::test_slim_generation
+```
+
+
+To restrict to specific recipes in the examples dir, the syntax is:
+
+```
+python3 -m pytest -k nonWF_recipe
```
petrelharp, line 24: very nice. But, what is `nonWF_recipe`? There aren't any example slim scripts named this?

hyanwong: Seems to be to me: https://github.com/tskit-dev/pyslim/pull/131/files#diff-5f283adae7de0cdcc50a5773f9873d8b62ec0d5e0313f6bdbda6ec4157671dbd

hyanwong: I guess it's obvious that these can be combined?

`python3 -m pytest tests/test_tree_sequence.py::test_slim_generation -k nonWF_recipe`

petrelharp: Ah, ok. But, that's `recipe_nonWF`, not `nonWF_recipe`.

petrelharp: oh wait maybe you renamed all the recipes?

hyanwong: I had to change the names and put `_recipe` at the end, not `recipe_` at the front (discussed in the initial comment).

petrelharp: What's the motivation for that? I thought you were removing the repetitive `recipe_` at the start... but it's at the end, now?

tests/slim_recipe_specs.py
```diff
@@ -0,0 +1,51 @@
...
+        value = {t: True, "everyone": True, "pedigree": True}
+        if s == 'early':
+            value['remembered_early'] = True
+        basic_slim_recipe_specs[f'record_everyone_{t}_{s}_recipe'] = value
+
+
+# These SLiM scripts read in an existing trees file; the "input" gives a .trees files produced
+# by recipes above that is appropriate for starting this script.
+restarted_slim_recipe_specs = {
+    "nucleotides_restart":   {"WF": True, "nucleotides": True, "no_op": True, "input": f"tests/examples/recipe_nucleotides.trees"},
```
petrelharp, line 42: this `input` looks different than the others

tests/README.md
hyanwong: The `-k` flag matches its arg anywhere in the test string. Since there were some recipes named `recipe_nonWF.slim` and `recipe_nonWF_early.slim`, then running `pytest -k recipe_nonWF` would run both. If you put `_recipe` on the end then that serves as a useful terminator to stop this.

hyanwong: I guess the other way would be to require the `.slim` prefix to be passed: `pytest -k recipe_nonWF.slim`, and then I would have to cut off the `.slim` before checking for the settings in `slim_recipe_specs.py`. Perhaps you prefer this? It's a bit more fiddly though.

tests/slim_recipe_specs.py
hyanwong: Could be. I haven't tested this yet. I just want to get some feedback as to whether the general approach was OK.

tests/README.md
hyanwong: Hmm, I forgot that the prefix isn't unique either. So we have `WF_early_recipe` and `nonWF_early_recipe`, both of which will now match with `-k WF_early_recipe`. Argh. Looks like we need `recipe_` back on the front and `.slim` on the end too.

tests/conftest.py
```diff
@@ -2,46 +2,120 @@
...
+        with FileLock(str(out_dir) + ".lock"):
+            # get the temp directory shared by all workers, so that simulation files
+            # are shared between workers
+            if out_dir.is_dir():
+                outfiles = slim_outfiles(out_dir)
+            else:
+                os.makedirs(out_dir, exist_ok=True)
+                outfiles = run_slim(recipe=recipe_name, out_dir=out_dir)
+    
+    ret = {o.key: o.post_process(o.path) for o in outfiles if os.path.isfile(o.path)}
```
petrelharp, line 119: very slick!

tests/README.md
petrelharp: arg, that's fiddly. But it makes it a bit more clear to have `.slim` on the end anyhow, IMO.

petrelharp:
Gee, this is great! Looks good. (And, sorry I got confused about the file names before I read your whole comment. :sheepish:)

> There's a FIXME because I'm not sure where in the code we should normally set the random seed if these tests are all parallelized.

I'm not sure either.

No other concerns! I do wonder if passing in the TREES_FILE and PEDIGREE_FILE is unnecessary complexity, since we could just pass in `OUTDIR` instead, but no big deal.

hyanwong:
> Gee, this is great! Looks good. (And, sorry I got confused about the file names before I read your whole comment. :sheepish:)

No probs. I'll try to get the whole thing working then.

> > There's a FIXME because I'm not sure where in the code we should normally set the random seed if these tests are all parallelized.
> 
> I'm not sure either.

I'll leave it as a FIXME for the time being then, and maybe ask Ben.

> No other concerns! I do wonder if passing in the TREES_FILE and PEDIGREE_FILE is unnecessary complexity, since we could just pass in `OUTDIR` instead, but no big deal.

If we explicitly pass the full file names, we only define them in one place. Otherwise we need to decide on a filename (e.g. `out.trees`) in SLiM, and redefine the same names in the python code too (and make sure they keep in sync). And we would also need to do the filename munging in SLiM. And the code to create the dir+filename in SLiM would be repeated in all the recipes, so I suspect it would result in more code duplication.

petrelharp:
Oh - about the seed - I think I do know the answer. The test set-up doesn't do any random stuff, so it doesn't need a seed set. And all the randomness we do *in* tests should have the seed specified, like
```
def test_whatever( ):
  random.seed(1234)
  a = random.sample(x)
  ...
```
... as we do in tskit. So no need to put it there. (It's possible that some of my tests don't set hte seed, but they should.)

hyanwong:
Hoorah! The tests all seem to be passing now. You can fully parallelise all the tests & recipes now, SLiM simulation output is now all saved to tmp dirs (but only run once each, even when multi-threaded, I hope), and there's a slick way to call the restarted tests while making sure that the sims on which they depend have been run:

```
    @pytest.mark.parametrize(
        'restart_name, basic_recipe', restarted_recipe_eq("no_op"), indirect=["basic_recipe"])
```

Then use the `basic_recipe["ts"]` which you are now sure is available, and run SLiM on the output. The fancy stuff is all done in `restarted_recipe_eq()` at the bottom of recipe_specs.py. I can explain how it works if it's not clear.

Weirdly the `restart_and_run_WF.slim` and `restart_and_run_nonWF.slim` seem not to have slim recipes in tests/examples, so I've commented them out for the time being.

The pytest command will also now work from whatever dir you want to run it in.

Ready for final review, I reckon @petrelharp . Then I can try to move on to the "retain"ing PR.

hyanwong:
P.s. there's not much left in `__init__.py` - it's mostly been moved to `conftest.py`. In fact, some of the stuff in `__init__.py` like `assertTablesEqual` can probably be swapped for functions which are now built in to tskit (I think?)

tests/examples/recipe_nonWF.slim
```diff
@@ -1,8 +1,8 @@
 initialize()
 {
-    setSeed(23);
-    defineConstant("OUTFILE", "recipe_nonWF.trees");
-    defineConstant("PEDIGREE", paste0(c(OUTFILE, ".pedigree")));
+    // OUTPREFIX can be defined on the command line to save in a different dir
```
petrelharp, line 3: whoops!

petrelharp: hm, and all the other recipes have the `setSeed(23)` line (which I think is a good idea) - should put that back in

tests/recipe_specs.py
```diff
@@ -0,0 +1,74 @@
...
+            k for k, v in basic_recipe_specs.items()
+            if exclude_if_has_key not in v and all(kk in v for kk in keys)
+        )
+
+# These SLiM scripts read in an existing trees file; the "input" gives a key in the
+# basic_recipe_specs array that will produce a "ts" file suitable for input
+restarted_recipe_specs = {
+    "restart_nucleotides.slim":   {"WF": True, "nucleotides": True, "no_op": True, "input": "recipe_nucleotides.slim"},
+    #"restart_and_run_WF.slim":    {"WF": True, "input": "recipe_init_mutated.slim"},
+    #"restart_and_run_nonWF.slim": {"nonWF": True, "input": "recipe_init_mutated.slim"},
```
petrelharp, line 57: hm, looks like these were just not working/being used, right?

tests/test_annotation.py
```diff
@@ -107,8 +105,8 @@ def verify_slim_restart_equality(self, in_ts, out_ts):
         out_tables.sort()
         self.assertTableCollectionsEqual(in_tables, out_tables, skip_provenance=-1)
 
-    def test_annotate_errors(self):
-        for ts in self.get_msprime_examples():
+    def test_annotate_errors(self, helper_functions):
+        for ts in helper_functions.get_msprime_examples():
```
petrelharp, line 109: hah, nice move here.

tests/test_annotation.py
```diff
@@ -288,66 +287,84 @@ def test_annotate_mutations(self):
...
 
 
 class TestReload(tests.PyslimTestCase):
     '''
     Tests for basic things related to reloading with SLiM
     '''
-
-    def test_load_without_provenance(self):
+    @pytest.mark.parametrize(
+        'restart_name, basic_recipe', restarted_recipe_eq("no_op"), indirect=["basic_recipe"])
```
petrelharp, line 342: So, what's `indirect` do? Makes it cached? [The docs](https://docs.pytest.org/en/stable/reference.html#pytest.python.Metafunc.parametrize) are kinda cryptic...

tests/recipe_specs.py
petrelharp: as you said above... I've made a note of this: https://github.com/tskit-dev/pyslim/issues/133

tests/test_metadata.py
```diff
@@ -71,97 +65,98 @@ def validate_slim_metadata(self, t):
...
+        assert tables.metadata['SLiM']['nucleotide_based'] == True
+
+
+
+    @pytest.mark.parametrize('basic_recipe', basic_recipe_eq("WF"), indirect=True)    
+    def test_WF_model_type(self, basic_recipe):
+        self.validate_model_type(basic_recipe["ts"], "WF")
+        
+    @pytest.mark.parametrize('basic_recipe', basic_recipe_eq("nonWF"), indirect=True)    
+    def test_WF_model_type(self, basic_recipe):
```
petrelharp, line 129: whoops! this is supposed to be `test_nonWF_model_type`

petrelharp:
Wow, that was a lot of work! And, it *works*!!  Thanks for the wonderful solutions, not to mention the cleaning up after me in a few places. I have been over this with a fairly coarse comb, and have muddled through these fixture things enough to convince myself it's doing all the stuff it did before. (although I may need you to explain how it all works sometime soon, while it's fresh...)  I only found two very minor changes in `tests/examples/recipe_nonWF.slim`, and one multiply-defined test; otherwise I'm happy to merge. 

(And, I've done a diff of the the tests run before and after this PR, and other than that multiply-defined test, they're all being run!)

tests/examples/recipe_nonWF.slim
hyanwong: Oops. Slippery fingers. Will reinstate.

tests/test_annotation.py
hyanwong: I think it feeds the parameter value through the fixture function. So passing `("restart_WF.slim", "recipe_WF.slim")` to `"restart_name, basic_recipe", ..., indirect=["basic_recipe"]` will call `basic_recipe("recipe_WF.slim")` and return that as the `basic_recipe` value in the test, whereas the `restart_name` parameter in the test function just contains the plain value `"restart_WF.slim"`

hyanwong:
> Wow, that was a lot of work! And, it _works_!! 

🎉 

>  I may need you to explain how it all works sometime soon, while it's fresh...

Yeah, it's not immediately obvious, sadly. But I think that's OK, the intention is clear to a reader.

> I only found two very minor changes in `tests/examples/recipe_nonWF.slim`, and one multiply-defined test; otherwise I'm happy to merge.

Done. Merge away!

> (And, I've done a diff of the the tests run before and after this PR, and other than that multiply-defined test, they're all being run!)

Great. How do you do that, by the way?

tests/test_annotation.py
petrelharp: oh!! gotcha.

petrelharp:
Yay!

I did
```
python3 -m pytest tests -n 1 -v &> temp_old
git checkout refactor-tests
python3 -m pytest tests -n 1 -v &> temp_new
diff <(cat temp_old | grep "^tests" | sort | uniq) <(cat temp_new | grep "^tests" | sed -e 's/\[.*\]//' | sort | uniq)
```
It's not checking whether the tests are being run on exactly the same scripts, but all the tests are being run, and none are being run on the *wrong* scripts, so I think it's all good.

PR #144, opened 2021-02-20
removed discrete msprime hack: closes #141

### LINKED ISSUE #141: Release that works with current msprime 1.0 alpha
The currently released version of pyslim doesn't work with msprime alphas because it's using  APIs that didn't make it into the actual release.

The simplest way to fix the code, I think, is to do 
```
recombination_map = msprime.RecombinationMap(
                   positions = [0.0, self.sequence_length],
                   rates = [recombination_rate, 0.0],
                   num_loci = int(self.sequence_length))
```

in all cases, and to get rid of the ``discrete_msprime`` thing [here](https://github.com/tskit-dev/pyslim/blob/4fa846bea0c15e9c06cec943671fdf5291636e29/pyslim/slim_tree_sequence.py#L445)

Msprime 1.0 should be fully compatible with code doing discrete simulations the old-school way, so this workaround isn't needed any more.

It would be good to get this released soon, so that pyslim users code won't break when msprime 1.0 is released. It's also stopping us from getting the stdpopsim test suite working with msprime 1.

petrelharp:
As suggested in #141 - should at least get stuff working.

### DIFF
--- a/pyslim/slim_tree_sequence.py
+++ b/pyslim/slim_tree_sequence.py
@@ -443,21 +443,10 @@ class SlimTreeSequence(tskit.TreeSequence):
             raise ValueError("The keep_first_generation argument is deprecated:"
                              "the FIRST_GEN flag is no longer used.")
 
-        # toggle for hacks below to deal with old msprime
-        discrete_msprime = hasattr(msprime, "RateMap")
-
-        if recombination_rate is not None:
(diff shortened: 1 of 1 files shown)


petrelharp:
Ack, this was wrong - I still need to check if RecombinationMap is passed in, at least.

PR #150, opened 2021-02-23
test on release slim also

petrelharp:
Run github CI tests both on SLiM master and release version.


petrelharp:
Oh, this doesn't work, duh.

PR #154, opened 2021-03-09
Switch docs to jupyterbook and update to msprime v1 in the process

petrelharp:
This supercedes most of the work in #101.
```
git fetch origin pull/154/head:new_docs
git checkout new_docs
cd docs
make
```
and then point your browser at `docs/_build/html`.


petrelharp:
TODO: remove `add_ancestral`

PR #174, opened 2021-06-10
method to convert alleles to nucleotides

### LINKED ISSUE #73: output VCF with legal REF/ALT columns
Currently we write out the ancestral state and derived state in the REF and ALT columns of the VCF file. These can be anything, for instance (from SLiM):
```
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	tsk_0	tsk_1	tsk_2	tsk_3	tsk_4	tsk_5	tsk_6	tsk_7	tsk_8	tsk_9
1	735760	.		16823049	.	PASS	.	GT	1|0	1|0	0|1	0|1	1|1	0|0	0|1	1|1	0|1	1|1
1	735786	.		0	.	PASS	.	GT	1|1	1|1	1|1	1|1	1|1	1|1	1|1	1|1	1|1	1|1
1	878418	.		42,956113,42	.	PASS	.	GT	2|2	2|2	2|2	2|2	2|2	2|2	2|2	2|2	2|2	2|2
```
 Unfortunately, (IIRC) the standard for VCF files says that REF and ALT need to be strings of nucleotides:
vcf

This was brought up [here](https://github.com/tskit-dev/pyslim/issues/65) as well, but more investigation by @attrna found a number of VCF parsing software that can't deal with it (eg picard, gatk, vcftools).

I'm not sure what to do here, really. Suggestions?

### LINKED ISSUE #168: provide method to remap alleles (eg for VCF output)
Sometimes we'd like to be able to change what the ancestral/derived states are, for instance, for doing VCF output of a SLiM-produced tree sequence. In tskit-dev/tskit#1081 we decided that the way to do this is to provide a method that remaps alleles. It's pretty easy to do this by hand now, so I'm not sure at what level of generality the API should work. In tskit-dev/tskit#1081 I proposed [having the method take a callable that takes in a Variant and return a list of new alleles](https://github.com/tskit-dev/tskit/pull/1081/files#diff-ecbb802115b51a593926a3015f4ff2ec444fa4dc4915af74f4959abc60c84df8R4747). This is on the one hand probably too general - we should be trying to make the common use case of converting things to nucleotides very easy - and not completely general - since distinct mutations can have the same derived allele, and this wouldn't let us remap those to different alleles.

For the SLiM-to-VCF problem, I think we want to do something like this, to get unique alleles for each mutation:
```python
import tskit, msprime
import random

ts = msprime.sim_ancestry(10, recombination_rate=0.01, sequence_length=100)
ts = msprime.sim_mutations(ts, rate=0.01, model=msprime.SLiMMutationModel(type=0))
t = ts.tables
t.sites.clear()
t.mutations.clear()

def make_alleles(n):
    nucs = ['A', 'C', 'G', 'T']
    alleles = nucs.copy()
    random.shuffle(alleles)
    while len(alleles) < n:
        new = [a + b for a in alleles for b in nucs]
        random.shuffle(new)
        alleles.extend(new)
    return alleles[:n]

for s in ts.sites():
    alleles = make_alleles(len(s.mutations) + 1)
    t.sites.append(s.replace(ancestral_state=alleles[0]))
    for j, m in enumerate(s.mutations):
        t.mutations.append(m.replace(derived_state=alleles[j+1]))

mts = t.tree_sequence()
for h in mts.haplotypes():
    print(h)
```

petrelharp:
Primary use case would be VCF output: if `ts` was produced by a SLiM nucleotide simulation then we'd do:
```
nts = pyslim.convert_alleles(ts)
with open('out.vcf', 'w') as f:
  nts.write_vcf(f)
```
If the simulation was produced by a non-nucleotide simulation, or a nucleotide simulation with some non-nucleotide mutations, then we'd do
```
nts = pyslim.convert_alleles(
            pyslim.generate_nucleotides(ts)
)
with open('out.vcf', 'w') as f:
  nts.write_vcf(f)
```

Closes #73. Closes #168.


pyslim/methods.py
```diff
@@ -84,3 +86,93 @@ def recapitate(ts,
     return SlimTreeSequence(recap, reference_sequence=ts.reference_sequence)
 
 
+def convert_alleles(ts, generate=True, seed=None):
```
jeromekelleher, line 89: I would have thought it would be clearer to have two functions ``convert_alleles`` which expects a SLiM nucleotide model and ``generate_alleles`` which will generate random alleles. It took a few goes at reading to docstring to get the current semantics (why would you not use the SLiM nucleotides, if they exist already?)

jeromekelleher (commented):
LGTM, but I think it would be easier to document and write if it was split into two functions.

bhaller:
I haven't looked at this closely.  I guess my first-blush question would be: if this remapping is primarily for the purpose of handling SLiM output, why not follow the same policy that SLiM follows, for consistency?  (That policy is outlined in detail in the manual.)  I'm not saying that policy is *good*; but if one is trying to wedge SLiM's world of infinite alleles and stacked mutations into VCF's world of nucleotide-based alleles, I'm not sure any policy is *good*, and if people want VCF output that really makes total sense, they should probably just run a nucleotide-based model in the first place.  Consistency seems like the most important goal, maybe?  So having SLiM's VCF output and pyslim's VCF output doing totally different things just seems potentially weird and confusing.  (On the other hand, I guess it gives users a choice of which style of VCF output they would prefer :->).

pyslim/methods.py
petrelharp: Hm, that's a nice idea. Right now it's using nucleotides when they exist and generating them if they don't, so it works with mixed nuc- and non-nuc mutations. But there's no need to make it so complex.

petrelharp:
> why not follow the same policy that SLiM follows, for consistency?

Good idea. But after reading the policy careful, I think my answer is that (a) that'd be way too much work to duplicate (since I couldn't just use `tskit.write_vcf`), and (b) I think it'd be nice to have an alternative for how multiallelic states are handled. SLiM for non-nucleotide models represents *all* mutations as A->T, with one line per mutation, and you say that there are only four nucleotides so you can represent at most only four states; but I think that's not right - you're allowed multi-nucleotide alleles in VCF (e.g., `A,CCA,TTTTT` in the REF column). Here, I'm generating unique alleles by using multicharacter strings. I think my goal here is to produce a legal VCF file (so, without duplicated POS rows) with as much information as possible (without dropping multiallelic sites), and in a relatively simple way.

petrelharp:
Ok, revised proposal:
```
convert_alleles(ts):
    ```
    Copies the nucleotide state of each SLiM mutation to the derived state, and the reference sequence to the ancestral state.
    Mutations with no nucleotide (i.e., -1 in metadata) will have a derived state of `"."` (i.e., "missing") and tree sequences without
    a reference sequence will have ancestral states of `"."`.
    ```

generate_nucleotides(ts, reference_sequence=None, keep=True):
    ```
    Generates random nucleotides, by choosing a uniformly random nucleotide that differs from the parental state,
    and (if `ancestral_sequence` is True) a random reference sequence. Nucleotide state is stored in mutation metadata;
    this method will modify the metadata and keep everything else the same. If `keep=True`, then mutations that already have
    a nucleotide state (i.e., not set to -1 in metadata) will retain their state.
    ```
```

bhaller:
> > why not follow the same policy that SLiM follows, for consistency?
> 
> Good idea. But after reading the policy careful, I think my answer is that (a) that'd be way too much work to duplicate (since I couldn't just use `tskit.write_vcf`), and (b) I think it'd be nice to have an alternative for how multiallelic states are handled. SLiM for non-nucleotide models represents _all_ mutations as A->T, with one line per mutation, and you say that there are only four nucleotides so you can represent at most only four states; but I think that's not right - you're allowed multi-nucleotide alleles in VCF (e.g., `A,CCA,TTTTT` in the REF column). Here, I'm generating unique alleles by using multicharacter strings. I think my goal here is to produce a legal VCF file (so, without duplicated POS rows) with as much information as possible (without dropping multiallelic sites), and in a relatively simple way.

OK, that's reasonable.  I think it might turn out to be pretty weird, for many users, that (conceptually) point mutations get represented as indels in the VCF.  I can see that maybe raising issues with downstream analysis tools, which is usually what people want VCF for in the first place.  But maybe not any weirder than how SLiM handles them, and as you say, at least it's fully VCF-compliant.  And people can import back into SLiM and write VCF from there if they really want that style of output.

Re: your revised proposal.  Hmm, "uniformly random nucleotide that differs from the parental state, and (if `ancestral_sequence` is True) a random reference sequence".  If there are multiple segregating mutations at a single site, is there any guarantee that those mutations will receive different random nucleotides?  Is that a concern?

petrelharp:
Sorry, I didn't make it clear - this revised proposal ditches the idea of using multi-nucleotides to ensure distinct alleles. I could provide that as an example somewhere, instead. In the current proposal, `generate_nucleotides` serves two purposes: (a) getting a tree sequence ready to load into SLiM for a nucleotide model, and (b) letting users output vcf files from any tree sequences. More specific requirements for VCF output (eg having all alleles distinct) could be done ad-hoc by users (and I'd provide an example).

bhaller:
> Sorry, I didn't make it clear - this revised proposal ditches the idea of using multi-nucleotides to ensure distinct alleles. I could provide that as an example somewhere, instead.

Ah, OK.  I think that's for the best; creating fictional indels just for VCF output seems rather extreme.

> In the current proposal, `generate_nucleotides` serves two purposes: (a) getting a tree sequence ready to load into SLiM for a nucleotide model, and (b) letting users output vcf files from any tree sequences. More specific requirements for VCF output (eg having all alleles distinct) could be done ad-hoc by users (and I'd provide an example).

I see, OK.  Makes sense as far as I can see, but I can tell you that people who use VCF files for their analysis have very specific/fussy requirements, so you might want to run these plans past an empirical biologist or two for a sanity check.  :->

petrelharp:
Well, this turned out to be unavoidably complicated, since substitutions will manifest as stacked mutations, and for both of these methods we need to figure out which SLiM mutation was the 'most recent' one, which is not straightforward. I'm deciding the most recent one is the one that
1. has the largest `slim_time`
2. is not present in the list of SLiM mutations for the parent mutation, if any
3. has the largest SLiM ID, if the other two fail to distinguish.

This is implemented [here](https://github.com/tskit-dev/pyslim/pull/174/files#diff-5197d510772af9508d710a5d4ec3471fe013a866e5641c4395c8170a63852258R123) and tested [here](https://github.com/tskit-dev/pyslim/pull/174/files#diff-322b5e720295c45c8a459d6cb762209ca5a3bc09873634731fbd5f8f9287c4eaR759).  It's not pretty.

@bhaller, mind having a look at whether the API and documentation makes sense (so, [here](https://github.com/tskit-dev/pyslim/pull/174/files#diff-5197d510772af9508d710a5d4ec3471fe013a866e5641c4395c8170a63852258R89) and [here](https://github.com/tskit-dev/pyslim/pull/174/files#diff-5197d510772af9508d710a5d4ec3471fe013a866e5641c4395c8170a63852258R164)) and whether what I said above seems right to you?  I suppose I could also be testing for agreement with what SLiM thinks is the nucleotides, by loading things into SLiM, but I have not.

bhaller:
Yeah, this is a pain for sure.  It would be nice if SLiM put the derived state in order, but the reason it doesn't is that it would take time, and thus slow down every simulation, and most simulations don't care about this; so it would slow down the many for a benefit to the few or the one, which would contradict Spock's dictum.  So.  I'll have a look at this tomorrow.

bhaller:
Hey.  OK, well, I looked at the implementation and test code, but it is beyond my rudimentary skills in Python; I really can't say whether what it's doing looks correct or not.  I'm just not a Python-fluent programmer, sorry.  Maybe some day, but since I use it quite infrequently, I've even forgotten most of the Python that I learned in the past.

I looked at the doc links you provided.  The doc for `convert_alleles` is very clear, seems good.  The one thing that gave me pause was this statement: "This method will produce an error... if any mutations do not have nucleotides".  Hmm.  It is legal and valid to do a nucleotide-based model that contains some non-nucleotide-based mutations; this can be useful, for example, if you want to use marker mutations to designate which genomes contain an inversion at a particular position, for example.  So, is there something better than throwing an error that this function could do when confronted by such tree sequences?

Also, a nit: your use of the word "insert" doesn't seem quite right, as "insert" means, to me, "add a new thing somewhere internally, while leaving what was there before unchanged"; you seem to mean "replace", not "insert", especially in "will instead insert the nucleotide from the mutation's metadata".

Another nit: "this method tries to assign" should start with a capitol letter.

Should `reference_sequence` be a string of letters, "ACGTACGT", or a string of integers, "01230123"?  I would have assumed letters, but then you discuss integer nucleotide values in a couple spots, so then I start to wonder.  Also: should the code check that `reference_sequence` is actually well-formed, containing what it is expected to contain, beyond just its length?  What happens if someone passes "Thequickbrownfoxjumpsoverthelazydog" for a reference sequence?  :->

That's all that I see.

petrelharp:
Perfect, thanks! And, sorry the python is so impenetrable.

>  It is legal and valid to do a nucleotide-based model that contains some non-nucleotide-based mutations; this can be useful, for example, if you want to use marker mutations to designate which genomes contain an inversion at a particular position, for example. So, is there something better than throwing an error that this function could do when confronted by such tree sequences?

Well, there might be something better; but I'm trying to keep things reasonly simple; so in this case I would recommend calling `generate_nucleotides` first (with `keep=True`, the default); this will insert nucleotides for the non-nucleotide mutations. I cannot think of a better general-purpose default; for more complex things the user will need to do it themselves.

TODO:
- [ ] `your use of the word "insert" doesn't seem quite right, as "insert" means, to me, "add a new thing somewhere internally, while leaving what was there before unchanged"; you seem to mean "replace", not "insert", especially in "will instead insert the nucleotide from the mutation's metadata".`
- [ ] ` "this method tries to assign" should start with a capitol letter.`
- [ ] clarify that `reference_sequence` should be a string of nucleotides (not digits).

And, good thought, but I don't think it's the job of this function to go checking all possible aspects of the reference sequence.

bhaller:
> Perfect, thanks! And, sorry the python is so impenetrable.

Just not a language I'm fluent in, I'm sure there's nothing wrong with the code.  :->

> > It is legal and valid to do a nucleotide-based model that contains some non-nucleotide-based mutations; this can be useful, for example, if you want to use marker mutations to designate which genomes contain an inversion at a particular position, for example. So, is there something better than throwing an error that this function could do when confronted by such tree sequences?
> 
> Well, there might be something better; but I'm trying to keep things reasonly simple; so in this case I would recommend calling `generate_nucleotides` first (with `keep=True`, the default); this will insert nucleotides for the non-nucleotide mutations. I cannot think of a better general-purpose default; for more complex things the user will need to do it themselves.

OK.  Perhaps the doc could suggest that workflow, rather than just saying that it produces an error?

> And, good thought, but I don't think it's the job of this function to go checking all possible aspects of the reference sequence.

OK.  In general I'm in favor of sanity-checking inputs for user-visible APIs whenever there is not a performance reason not to, so that the user gets a good error message instead of just mysteriously/confusingly wrong behavior, but it certainly does add more work.  Of course my example was a bit flip, but it might catch a legit typo in a hand-typed string like "ACGTAVGT", or a confused user doing "01230123", or FASTA data that contains extended nucleotide codes like U, R, Y, etc. (https://en.wikipedia.org/wiki/FASTA_format#Sequence_representation).  I have had users try to input such codes into SLiM for an ancestral sequence.  :->

petrelharp:
Compelling. Added these to the TODOs.

PR #181, opened 2021-07-06
Fix pickling of SlimTreeSequence

### LINKED ISSUE #180: Unpickling TreeSequence from pyslim doesn't initialize metadata, and raises exception
I've run into an unexpected exception with tskit 0.3.6 — a description and MRE are below. The zip file [tskit_mre.zip](https://github.com/tskit-dev/tskit/files/6767413/tskit_mre.zip) contains a SLiM simulation routine based on recipe 17.1 and an example tree sequence with metadata. Loading the tree using pyslim and trying to pickle and unpickle it as below, 

```python
import pyslim as psl
import tskit as tsk
import pickle

[x.__version__ for x in (psl, tsk)]
# ['0.600', '0.3.6']

tr = psl.load('./test.tree')
with open('test.pkl', 'wb') as f:
  pickle.dump(tr, file=f)

with open('test.pkl', 'rb') as f:
   tr2 = pickle.load(file=f)
```
leads to the following error,

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-11-dc9156f6d230> in <module>
      1 with open('../data/slim_sims/test.pkl', 'rb') as f:
----> 2     t = pickle.load(f)

~/miniconda3/envs/bioinfo/lib/python3.7/site-packages/tskit/trees.py in __setstate__(self, tc)
   3435 
   3436     def __setstate__(self, tc):
-> 3437         self.__init__(tc.tree_sequence().ll_tree_sequence)
   3438 
   3439     def __eq__(self, other):

~/miniconda3/envs/bioinfo/lib/python3.7/site-packages/pyslim/slim_tree_sequence.py in __init__(self, ts, reference_sequence, legacy_metadata)
    155     def __init__(self, ts, reference_sequence=None, legacy_metadata=False):
    156         self.legacy_metadata = legacy_metadata
--> 157         if not (isinstance(ts.metadata, dict) and 'SLiM' in ts.metadata
    158                 and ts.metadata['SLiM']['file_version'] in compatible_slim_file_versions):
    159             tables = ts.dump_tables()

AttributeError: '_tskit.TreeSequence' object has no attribute 'metadata'
```

This is with tskit 0.3.6 — on a different machine with 0.3.5, this works fine, so I suspect this is a regression from this change in how pickled tree sequences are initialized: https://github.com/tskit-dev/tskit/pull/1298/files#diff-ecbb802115b51a593926a3015f4ff2ec444fa4dc4915af74f4959abc60c84df8L3318. I would submit a PR, but I'm not sure the best way to initialize the object in a way that preserves the metadata.

benjeffery:
Fixes #180

### DIFF
--- a/pyslim/slim_tree_sequence.py
+++ b/pyslim/slim_tree_sequence.py
@@ -196,6 +196,20 @@ class SlimTreeSequence(tskit.TreeSequence):
         self.individual_populations[which_indiv] = ts.tables.nodes.population[has_indiv]
         self.individual_times[which_indiv] = ts.tables.nodes.time[has_indiv]
 
+    def __getstate__(self):
+        return {
+            'tables':self.dump_tables(),
+            'legacy_metadata': self.legacy_metadata,
+            'reference_sequence': self.reference_sequence
+        }
+
+    def __setstate__(self, state):
+        self.__init__(
+            state['tables'].tree_sequence(),
+            state['reference_sequence'],
+            state['legacy_metadata']
+        )
+
     @property
     def slim_generation(self):
         # return self.metadata['SLiM']['generation']
--- a/tests/test_tree_sequence.py
+++ b/tests/test_tree_sequence.py
@@ -1,6 +1,7 @@
 """
 Test cases for tree sequences.
 """
+import pickle
 import random
 import os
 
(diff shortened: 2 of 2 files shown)


petrelharp:
Um, I trust you that this is ok? I have totally no idea what's going on here, though. Want to explain?

Also, we might as well fix this, but the plan is to deprecate SlimTreeSequence in the next release.

jeromekelleher (approved):
Looks correct to me.

Because the constructor for SLiMTreeSequence has a different signature to the tskit version and extra stuff needs to be saved, the pickle machinery has to be overridden.

petrelharp (approved):

Issue #175, opened 2021-06-11
add example of getting the correct individuals


petrelharp:
The distinction between samples and individuals is still opaque. We could add an example to the start of the tutorial that 
1. runs a nonWF simulation
2. simplifies it
3. tries to get variants from all alive individuals, but gets an error
4. correctly checks which individuals are samples to get variants

PR #182, opened 2021-07-06
edited tutorial.md to take a sample of individuals

chriscrsmith:
Referencing issue [#175](https://github.com/tskit-dev/pyslim/issues/175), I offered some edits to the tutorial for (1) taking a sample from a population, (2) taking a sample after simplification.

### DIFF
--- a/pyslim/slim_tree_sequence.py
+++ b/pyslim/slim_tree_sequence.py
@@ -196,6 +196,20 @@ class SlimTreeSequence(tskit.TreeSequence):
         self.individual_populations[which_indiv] = ts.tables.nodes.population[has_indiv]
         self.individual_times[which_indiv] = ts.tables.nodes.time[has_indiv]
 
+    def __getstate__(self):
+        return {
+            'tables':self.dump_tables(),
+            'legacy_metadata': self.legacy_metadata,
+            'reference_sequence': self.reference_sequence
+        }
+
+    def __setstate__(self, state):
+        self.__init__(
+            state['tables'].tree_sequence(),
+            state['reference_sequence'],
+            state['legacy_metadata']
+        )
+
     @property
     def slim_generation(self):
         # return self.metadata['SLiM']['generation']
--- a/tests/test_tree_sequence.py
+++ b/tests/test_tree_sequence.py
@@ -1,6 +1,7 @@
 """
 Test cases for tree sequences.
 """
+import pickle
 import random
 import os
 
@@ -83,6 +84,15 @@ class TestSlimTreeSequence(tests.PyslimTestCase):
         # check persists through simplify
         simp = ts.simplify(ts.samples())
         assert simp.slim_generation == new_sg
+
+    def test_pickle(self):
+        ts = self.clean_example().tree_sequence()
+        ts = pyslim.SlimTreeSequence(ts)
+        roundtripped = pickle.loads(pickle.dumps(ts))
+        assert roundtripped == ts
+        # __eq__ doesn't check pyslim properties:
+        assert roundtripped.reference_sequence == ts.reference_sequence
+        assert roundtripped.legacy_metadata == ts.legacy_metadata
 
 
 class TestSlimTime(tests.PyslimTestCase):


docs/tutorial.md
```diff
@@ -394,7 +394,107 @@ What's going on here? Let's step through the code.
...
 (sec_extracting_individuals)=
 
-## Extracting particular SLiM individuals
+## Extracting SLiM individuals
+
+### Extracting a sample of individuals
+
+Often we would like to sample a number of individuals from a population.
+This section provides an example workflow and draws
+attention to some easy-to-miss details for this approach.
```
petrelharp, line 403:
```suggestion
attention to some common errors.
```

docs/tutorial.md
```diff
@@ -394,7 +394,107 @@ What's going on here? Let's step through the code.
...
+
+Let's start with a nonWF simulation similar to what we've seen previously.
```
petrelharp, line 405: Is this the *same* as in `example_sim.slim`?

docs/tutorial.md
```diff
@@ -394,7 +394,107 @@ What's going on here? Let's step through the code.
...
+```
+
+```{code-cell}
+%%bash
+slim -s 1 sample.slim
+```
+
+The first, most common method to extract individuals is simply to get all
+those that were alive at a particular time,
+using {meth}`SlimTreeSequence.individuals_alive_at`. For instance, to get
```
petrelharp, line 417:
```suggestion
using {meth}`.SlimTreeSequence.individuals_alive_at`. For instance, to get
```

docs/tutorial.md
```diff
@@ -394,7 +394,107 @@ What's going on here? Let's step through the code.
...
+ts = pyslim.SlimTreeSequence(msprime.mutate(orig_ts, rate=1e-8, random_seed=1))
+with open("example_snps.vcf", "w") as vcffile:
+    ts.write_vcf(vcffile, individuals=keep_indivs)
+```
+
+
+### Extracting individuals after simplification
+
+In cases where we want to obtain genotypes *after* simplification, we
+need to pay attention to details concerning individual nodes. As
```
petrelharp, line 443: I think we should avoid saying "individual nodes", since it confuses "individuals" and "nodes", which are different. What did you mean here?

docs/tutorial.md
```diff
@@ -394,7 +394,107 @@ What's going on here? Let's step through the code.
...
+with open("example_snps.vcf", "w") as vcffile:
+    ts.write_vcf(vcffile, individuals=indivlist)
+```
+
+:::{note}
+The {meth}`.Node.is_sample` check is a necessary and easy to
+forget step. If this step is omitted, the program will attempt
+to extract genotypes for any isolated nodes, e.g. non-sample
+nodes that were part of the simplified tree.
+This will result in an error message.
```
petrelharp, line 477: What do you think about actually demonstrating the error message? I'm not sure if it's good practice - on the one hand, people can more easily find the error message they might get; on the other hand, it's dangerous to put bad code where someone might copy-paste it.

docs/tutorial.md
```diff
@@ -394,7 +394,107 @@ What's going on here? Let's step through the code.
...
+    ind = ts.individual(i)
+    if ts.node(ind.nodes[0]).is_sample():                                                                              
+       indivlist.append(i)
+with open("example_snps.vcf", "w") as vcffile:
+    ts.write_vcf(vcffile, individuals=indivlist)
+```
+
+:::{note}
+The {meth}`.Node.is_sample` check is a necessary and easy to
+forget step. If this step is omitted, the program will attempt
```
petrelharp, line 474: We should probably explain why it's necessary a bit more explicitly.

docs/tutorial.md
```diff
@@ -394,7 +394,107 @@ What's going on here? Let's step through the code.
...
+This will result in an error message.
+:::
+
+
+The above code uses {meth}`.Node.is_sample` to check whether a node
+was part of our sample. This information was saved during the
+simplification step, and is saved in the ``flags`` attribute of the
+node metadata. For example, below shows the first node
+from two individuals: one individual with a sample
+flag (Individual 0) and one non-sample individual (Individual 2).
```
petrelharp, line 486: I'm not so sure about whether this goes here - the purpose is demonstrating how to interpret the `flags` attribute. So it should either say explicitly that `tskit.NODE_IS_SAMPLE` is `1`, and so if `node.flags & tskit.NODE_IS_SAMPLE` is nonzero then it's a sample... but this gets into explaining bitwise flags, and this might not be the right place for it.

petrelharp:
Looks good, thanks! A couple of questions.

Also, FYI - it's more usual to do a PR from a separate branch; that will make it so you won't have to force-push  to `main` on your fork later.

docs/tutorial.md
chriscrsmith: Yes. So, should delete the simulation step

docs/tutorial.md
chriscrsmith: good point. I meant specific nodes

docs/tutorial.md
chriscrsmith: I remember you suggesting that and I think it makes sense, showing the actual error message. I'm familiar with the error-True parameter in Rmarkdown but I couldn't figure out how to show the error, here. Do you know how?

chriscrsmith:
Thanks for looking over this. 

I'll revise- in another branch- and do a new PR. Sound good?

docs/tutorial.md
```diff
@@ -394,7 +394,87 @@ What's going on here? Let's step through the code.
...
+ts = pyslim.SlimTreeSequence(msprime.mutate(orig_ts, rate=1e-8, random_seed=1))
+with open("example_snps.vcf", "w") as vcffile:
+    ts.write_vcf(vcffile, individuals=keep_indivs)
+```
+
+
+### Extracting individuals after simplification
+
+In cases where we want to obtain genotypes *after* simplification, we
+need to pay attention to specific node IDs. As
```
petrelharp, line 433: Isn't what we need the node IDs for the simplification step? Getting the genotypes of a subset we can do without simplification, and don't need the node IDs.

docs/tutorial.md
```diff
@@ -394,7 +394,87 @@ What's going on here? Let's step through the code.
...
+of nodes as input:
+
+```{code-cell}
+keep_nodes = []
+for i in keep_indivs:
+    keep_nodes.extend(orig_ts.individual(i).nodes)
+sts = rts.simplify(keep_nodes)
+```
+
+There are often some nodes retained
```
petrelharp, line 444: there are (almost) *always* nodes retained that are not samples. Do you mean that there are often non-sample *individuals*?

docs/tutorial.md
petrelharp: some sort of `try ... except` block?

chriscrsmith: cool, I added a block

docs/tutorial.md
chriscrsmith: Taking another shot at this paragraph:

The Node.is_sample() check is necessary, here. This is important to avoid taking genotypes for any non-sample individuals, i.e. individuals alive at the end of the simulation that had one or more nodes retained in the simplified tree sequence. In fact, we can’t rely on the above error message: sometimes there are alive, non-sample nodes that are not isolated, and these get added to the vcf along with the sample genotypes.

docs/tutorial.md
chriscrsmith: deleted this paragraph

docs/tutorial.md
chriscrsmith: Sorry I might not understand correctly. My thought process was, we can get genotypes for a subset without simplification and don't need the node IDs- that example is under "Extracting a sample of individuals" in my doc. But in cases where simplification _is_ used, then we risk extracting genotypes for isolated nodes, or individuals that weren't part of our intended sample (the node list given to simplify()). Am I getting this wrong?

What if I change the highlighted bit to:
        "If we simplify, we need to pay attention to specific node IDs. As mentioned earlier, SlimTreeSequence.simplify() takes a list of nodes as input:"

docs/tutorial.md
petrelharp: How about
```
The {meth}`.Node.is_sample` check is necessary here because there may be individuals alive at the end of the simulation
that are still present in the tree sequence after simplification
despite not being chosen as samples.
This could happen, for instance, if two siblings are in ``keep_indivs`` above but a still-living parent is not.
```
I think explaining the error message should come after the error, not in this `:::{note}`.

docs/tutorial.md
petrelharp: Ah ha - I see what you mean. Yes, sounds good.

PR #189, opened 2021-07-29
population_size method

### LINKED ISSUE #186: add method to compute discretized population density
Make a method to compute the population size in intervals of x, y, and time from a tree sequence. It should take a tree sequence, bins in the x and y directions, and bins in time and return an easy to plot 3-d array of the number of individuals alive in each x and y bin averaged across the times in each time bin.

```
def population_size(ts, x_bins = 1, y_bins = 1, time_bins = 1)
```

If `x_bins`, `y_bins`, or `time_bins` is a single number, the breakpoints will be `np.linspace(0, max(x), xbins + 1)`. If any of the bins is an array it will be the breakpoints.

Returns a 3-d array with dimensions (number of x breakpoints - 1) by (number of y breakpoints - 1) by (number of time breakpoints - 1)

The goal is to make it easy to make plots like this:
```
import matplotlib.pyplot as plt
x_bins = [0, 10, 20]
time_bins = np.linspace(0, 1000, 101)
popsize = pyslim.population_size(ts, x_bins = x_bins, time_bins = time_bins)
plt.pcolormesh(time_bins, x_bins, popsize)
```
image

petrelharp:
Closes #186.

### DIFF
--- a/pyslim/spatial.py
+++ b/pyslim/spatial.py
@@ -12,18 +12,32 @@ from .util import *
 
 def in_location_bin(locations, x0, x1, y0, y1):
     '''
-    Takes the locations of individuals and  tests if each is within [x0, x1), [y0, y1). 
-    Returns a vector the same length as locations with True if they are within the bounds and False otherwise
-    locations is a n x 3 array with rows individuals and [:, 0] the x coordinate and [:, 1] the y coordinate
-    Returns an n x 1 array, TRUE if individual is in the bin
-    Each row corresponds to an individual
+    Takes the locations of individuals and tests if each is within [x0, x1) and [y0, y1). Returns a list the same length as the number of individuals with True if the corresponding individual is within the bounds and False otherwise.
+    
+    :param array locations: An n x 3 array with rows individuals, the first column the x coordinates, and the second column the y coordinates.
+    :param float x0: The lower x coordinate boundary.
+    :param float x1: The upper x coordinate boundary.
+    :param float y0: The lower y coordinate boundary.
+    :param float y1: The upper y coordinate boundary.
     '''
     return(np.logical_and(np.logical_and(locations[:,0] < x1, locations[:,0] >= x0),
                np.logical_and(locations[:,1] < y1, locations[:,1] >= y0)))
 
 def average_time_alive(birth_times, death_times, t0, t1):
     '''
-    Finds average population size in [t0, t1) using time alive
+    Calculates the average population size in [`t0`, `t1`) from birth times and death times of individuals. Helper function for `population_size`. 
+
+    'Average population size' is the sum over all individuals of the length of time each individual was alive in [`t0`, `t1`), divided by `t1 - t0`. For integer t0 and t1, this is equivalent to the average over all integer time steps in [t0, t1) of the number of individuals alive at that time step.
+
+    Time goes bacwards, so t1 is further in the past than t0 and birth times are greater than death times. Birth times and death times should already include an offset for the stage that individuals are remembered at. See `population_size` code.
+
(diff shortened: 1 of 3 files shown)


petrelharp:
Todo:
- document
- make helper functions private

pyslim/spatial.py
```diff
@@ -3,61 +3,103 @@
...
 
-# probably don't need all of these
 from .slim_tree_sequence import *
-from .slim_metadata import *
-from .provenance import *
-from .util import *
 
 
-def in_location_bin(locations, x0, x1, y0, y1):
+def _in_location_bin(locations, x0, x1, y0, y1):
```
petrelharp, line 9: underscores at the start of the functions make it so this method won't be exported by pyslim - and, I don't think we want to be making this a public function; we just want to use it in this file. Same with average_time_alive.

pyslim/spatial.py
```diff
@@ -3,61 +3,103 @@
...
+    include an offset for the stage that individuals are remembered at. See
+    `population_size` code.
+
+    For example, suppose t0 = 1, t1 = 4, and there are four individuals with
+    birth times [3, 1, 5, 4] and death times [0, 1, 2, 4]. The first individual
+    is alive for three time steps in [t0, t1): 1, 2, and 3; the second
+    individual for one time step: 1; the third individual for two time steps: 2
+    and 3, and the fourth individual for zero (since the time interval is open
+    on the right and it was born at t1). So the average population size is (3 +
+    1 + 2 + 0)/(4-1) = 2.
```
petrelharp, line 58: this I just formatted so the lines weren't really long.

pyslim/spatial.py
```diff
@@ -3,61 +3,103 @@
...
+    returns a 3-d array with dimensions ``(nx, ny, nt)``.
+    The ``(i,j,k)``th element of the array is the average number of
+    individuals with ``x`` coordinate in the half-open interval ``[x_bins[i], x_bins[i + 1])``
+    and ``y`` coordinate in the half-open interval ``[y_bins[i], y_bins[i + 1])``,
+    averaged across all times in the half-open interval ``[time_bins[k], time_bins[k + 1])``.
+
+    For integer endpoints of the time bins, this average is equivalent to
+    recording the number of indivduals that are alive at each time in the time
+    interval and have location in the relevant location bin, then taking the
+    mean of these recorded population sizes.
```
petrelharp, line 92: This bit I edited to be simpler - see what you think?

pyslim/spatial.py
```diff
@@ -95,36 +137,30 @@ def population_size(ts, x_bins, y_bins, time_bins, stage='late', remembered_stag
...
-    # Initialize array to store popoluationsize
+
     nxbins = len(x_breaks) - 1
     nybins = len(y_breaks) - 1
     ntbins = len(time_breaks) - 1
     popsize = np.empty((nxbins, nybins, ntbins))
-    popsize_validate = np.empty((nxbins, nybins, ntbins))
-    #print(np.shape(popsize))
 
-    # Location, times, and ages of individuals
```
petrelharp: I deleted some of these comments that didn't seem to add much any more

pyslim/spatial.py
```diff
@@ -95,36 +137,30 @@ def population_size(ts, x_bins, y_bins, time_bins, stage='late', remembered_stag
...
     locations = ts.individual_locations
-    times = ts.individual_times
```
petrelharp: we never used tihs

pyslim/spatial.py
```diff
@@ -3,61 +3,103 @@
...
-def average_time_alive(birth_times, death_times, t0, t1):
+    return np.logical_and(
+            np.logical_and(
+                locations[:,0] < x1,
+                locations[:,0] >= x0),
+            np.logical_and(
+                locations[:,1] < y1,
+                locations[:,1] >= y0
+            )
+        )
```
petrelharp, line 32: this line was long, so I spread it out.

petrelharp:
Looks good! I just want to write one more test, with a known answer.

Also: do you know how fast this is for a big-ish tree sequence with a fine spatial grid (e.g., 100x100)?

petrelharp:
Ok, I've added a simple test to catch the edge cases; it works!  Have a look at my changes; see what you think.

petrelharp:
Approved by @giliapatterson - merging!!

PR #190, opened 2021-08-02
WIP Union vignette

mufernando:
WIP


docs/phylo_bgs.slim
```diff
@@ -0,0 +1,30 @@
+initialize()
+{
```
petrelharp, line 2: perhaps say up top what the things that need to be defined are? (`treefile`, `gens`, and `popsize`?)

docs/phylo_bgs.slim
```diff
@@ -0,0 +1,30 @@
...
+    initializeGenomicElementType("g1", m1, 0.1);
+    initializeGenomicElement(g1, 0, 1e4-1);
+    initializeRecombinationRate(1e-8);
+}
+
+1 early() {
+    // if no input tree sequence is provided, then start a subpopulation
+    if (treefile == "") {
+        sim.addSubpop("p1", popsize);
+        finalgen = gens;
```
petrelharp, line 16: `num_gens` might be nicer than `gens`

docs/vignette_parallel_phylo.md
```diff
@@ -0,0 +1,102 @@
...
+branch or not, read the previous tree sequence (if necessary) and change the
+population size accordingly.
+Note that when you read a tree sequence into SLiM, the generation counter will
+be updated with the time encoded in the tree sequence, so we need to set the end
+of the simulation as the length of the branch (`gens`) plus the total length of the
+loaded tree sequence.
+At the end of the simulation, we call `sim.treeSeqRememberIndividuals` right 
+before saving the resulting tree sequence. This is necessary because we need to
+ensure the individuals in the final generation are never dropped from the tree 
+sequence in simplification, as they will later be used to glue the tree sequences
```
petrelharp, line 67: maybe instead of `in simplification`, say `in future runs of SLiM which are started using the output of the simulation` or something

docs/phylo_bgs.slim
```diff
@@ -0,0 +1,34 @@
+// The following constants need to be defined:
+// - outfile: path to save the tree sequence of the simulation.
+// - popsize: population size.
+// - num_gens: run simulation for num_gens generations.
+// - infile: a string with path to tree sequence to start from; optional.
+initialize()
+{
+    initializeSLiMModelType("WF");
+    initializeTreeSeq();
+    initializeMutationRate(1e-6);
```
petrelharp, line 10: this rate is very high?

docs/phylo_bgs.slim
```diff
@@ -0,0 +1,34 @@
...
+    initializeMutationType("m1", 0.5, "f", -0.01);
+    initializeGenomicElementType("g1", m1, 0.1);
+    initializeGenomicElement(g1, 0, 1e4-1);
```
petrelharp, line 13: perhaps do a longer region and lower mutation rate?

docs/phylo_bgs.slim
```diff
@@ -0,0 +1,34 @@
...
+    initializeRecombinationRate(1e-8);
+}
+
+1 early() {
+    // if no input tree sequence is provided, then start a subpopulation
+    if (infile == "") {
```
petrelharp, line 19: alternatively, this could be
```
if (!exists("infile")) {
```
allowing the infile to be skipped entirely.

docs/vignette_parallel_phylo.md
```diff
@@ -0,0 +1,208 @@
...
+to check it out).
+The main idea here is that you can specify dependency between files and `make`
+works its magic to run the simulations in the right order.
+
+```{code-cell}
+f = open("sims.make", "w")
+print(f"all: {' '.join(df.outfile.to_list())}\n", file=f)
+for i, row in df.iterrows():
+    print(f"{row.outfile}: {row.infile}", file=f)
+    print(f"\tslim -d \"infile='{row.infile}'\" -d popsize={row.popsize} -d num_gens={row.edgelen} -d \"outfile='{row.child}.trees'\" phylo_bgs.slim\n", file=f)
```
petrelharp, line 95: I love it! We could instead do 
```
     print(f"\tslim -d \"infile='$<\" -d popsize={row.popsize} -d num_gens={row.edgelen} -d \"outfile='$@'\" phylo_bgs.slim\n", file=f)
```
... but I suppose that'd be less transparent.

docs/vignette_parallel_phylo.md
```diff
@@ -0,0 +1,208 @@
...
+%%bash
+cat sims.make
+```
+
+We can run make and specify the number of commands to be run simultaneously (using `-j`).
+
+```{code-cell}
+:tags: ["hide-output"]
+%%bash
+make -f sims.make -j 1
```
petrelharp, line 110: shouldn't this have `-j 3`?

petrelharp:
I made some minor tweaks - but, I see it's not quite done at the end? Looking for the elegant way to do that last recursive union?

mufernando:
> I made some minor tweaks - but, I see it's not quite done at the end? Looking for the elegant way to do that last recursive union?

Thanks! I did the union recursion starting [here](https://github.com/tskit-dev/pyslim/pull/190/files#diff-93d89df0bd2f18cffcd36ab4a3c071a1002f8eaa816f4b2473bd8792f47d90d1R182). We could do some sort of analysis at the end to show it worked, is that what you mean?

docs/vignette_parallel_phylo.md
mufernando: ah, yeah, totally forgot about those make wildcards. though I agree it would make this a bit harder to read.

docs/phylo_bgs.slim
mufernando: then it is trickier to deal with this in the print of the slim command.

docs/vignette_parallel_phylo.md
mufernando: fixed it

mufernando:
I think this is ready to be reviewed. I included two bits at the end that make use of the union-ed tree sequence:

- Count the number of samples in each of the independently simulated simulations and make sure that checks out
- Added a figure with some trees illustrating ILS (I made one of the branches shorter and with a lot of individuals

Still think we might need to work on the text a bit, maybe to make it more detailed?

docs/vignette_parallel_phylo.md
```diff
@@ -206,43 +227,85 @@ def union_recursion(df, focal="root", merged=None):
...
+    for i, _ in enumerate(pops)
+]
+subsample_nodes = [
+    np.concatenate([tsu.individual(i).nodes for i in x])
+    for x in subsample_indivs
+]
+tsus = pyslim.SlimTreeSequence(
+    tsu.simplify(
+        np.concatenate(subsample_nodes),
+        filter_populations=False,
```
mufernando, line 293: did this work without the keep_input_rooots? I noticed that we would lose the root nodes without it.

mufernando (commented):
this looks great, thanks!

docs/vignette_parallel_phylo.md
petrelharp: I recapitated earlier on instead - I figured that was less confusing.

petrelharp:
Ah, one more thing - I don't think we should be checking in the .trees files (although I notice that previous ones are, hm).

petrelharp:
... well, I guess we have the .trees files in for good reason. (CI breaks otherwise)

mufernando:
I think CI can't run make -- or at least not with more than one core.

petrelharp:
> I think CI can't run make -- or at least not with more than one core.

Ah, would that actually make the file not available when it's supposed to be? The test passes otherwise.

PR #200, opened 2021-09-12
fixup of union vignette

petrelharp:
So, I tried to fix up the problem with counting the number of samples in each population with the union vignette  - because it's an eas fix and even if that goes through soon, it still won't be available in tskit until the next release - but, in so doing discovered that the simulations weren't being run, and they don't seem to actually run on my computer. Currently I got the `Shared portions of the tree sequences are not equal.` error.

Do you know what's going on, @mufernando?  I think the vignette *should* actually run the sims (which we should make smaller if it takes a while to run).


mufernando:
Let me look into this. We can run the sims everytime, the only thing is we can't call make with more than one core, I think.

mufernando:
well, I just ran all the sims locally many times and never got that error. also I made some changes to this PR and was trying to push it back to you, but I don't have rights to your repo so I made a new PR just to make sure we don't get test errors, see #201.

mufernando:
FYI I have tskit 0.3.6 and SLiM 3.6 in case that matters for getting the union error.

PR #216, opened 2021-12-11
switch to new reference sequence

petrelharp:
WIP.


petrelharp:
These tests all pass *locally*, using `master` SLiM and msprime from [this PR](https://github.com/tskit-dev/msprime/pull/1945). I'm merging, so it doesn't get too big.

PR #230, opened 2021-12-17
deprecate SlimTreeSequence

petrelharp:
WIP.

TODO:
- deal with ts.indivdiual_locations
- remove pyslim.load( ) from docs

PR #231, opened 2021-12-17
halfway deprecations

petrelharp:
In the interest of getting out *a* release that works with stdpopsim soon, this is halfway to #230, and I think a good halfway point. I'd like to take this and release it as 0.700 (after checking it works with stdpopsim).


petrelharp:
Merging this.

PR #253, opened 2022-05-14
Slim v4

petrelharp:
Work in progress.


petrelharp:
TODO:

- [ ] switch "generation" for "tick" (/"cycle" when appropriate)
- [ ] add `"first"`
- [ ] remove `get_provenance( )` method

petrelharp:
This is timing out on docs build; I'll deal with that elsewhere.

PR #273, opened 2022-08-09
more testing of population remapping

### LINKED ISSUE #256: add tests of population remapping
See https://github.com/MesserLab/SLiM/issues/319: we'd like to test

- [x]     alive nodes with population not having SLim metadata produces an error
- [x]     reading of multiple tree sequences whose populations conflict produces an error
- [x]     reading of multiple tree sequences whose populations are remapped to not conflict works
- [x]     existing non-SLiM metadata in population table entries not referred to by alive nodes is not modified (including at the end of the table, past all slim populations)
- [x]     saving out of a tree sequence with an empty pop in nonWF still writes out the empty pop, and reloading it recovers that pop
- [x]     when remapping is used, carryover metadata gets shifted to the new remapped index unmodified, without errors. 
- [x]     a remapping that leads to a conflict between the remapped subpops correctly errors out
- [x]     when carryover metadata is present, SLiM correctly requires a remapping entry for that subpop index (if any remapping is being done at all)
- [x]     when a migration table is present, which SLiM does read in for WF models, remapping of the source_subpop for the migrations works correctly (which it definitely does not at the moment)

petrelharp:
This will close #256.


bhaller:
Awesome, thanks!  What's up with the multiple cancelled and one failed test in the checks above?  Is that something I should worry about, or no?

petrelharp:
I'll look into the failing tests shortly. The test for [referring to non-extant pops](https://github.com/tskit-dev/pyslim/pull/273/files#diff-20af20d49a99c5b9ba68e29a95fa313e4233f7d02d49c126a1ed04008356e1a9R660) is currently commented out.

petrelharp:
Well, the tests failed because I forgot to commit the recipe for some of them. =)

Issue #179, opened 2021-06-17
add method to get the largest SLiM mutation ID


petrelharp:
... so we know where to start mutating from

petrelharp:
... as is done [here](https://github.com/popsim-consortium/stdpopsim/blob/de8c2150b4f569b6a3ed1bd7d6d9d2e12b02ba82/stdpopsim/slim_engine.py#L1306) for instance.

bhaller:
Easy and useful, right?

petrelharp:
This is done in #290.

PR #290, opened 2022-08-12
max_slim_mutation_id function

mufernando:
Creates a function that returns the maximum SLiM mutation ID. Useful for adding more mutations with `msprime.sim_mutations()`.

Solves #179

### DIFF
--- a/pyslim/methods.py
+++ b/pyslim/methods.py
@@ -608,20 +608,24 @@ def annotate_tables(tables, model_type, tick, cycle=None, stage="early", referen
     if reference_sequence is not None:
         tables.reference_sequence.data = reference_sequence
 
-def max_slim_mutation_id(ts):
+def next_slim_mutation_id(ts):
     '''
-    Returns the largest SLiM mutation ID in this tree sequence. This is useful because
-    if you want to add more mutations to your tree sequence using :func:`msprime.sim_mutations`,
-    you may need to specify the parameter `next_id` in your 
-    :class:`msprime.SLiMMutationModel` to be larger than any existing mutation IDs.
-    Setting `next_id` equal to the output of this function + 1
-    will allow the mutated tree sequence to be read in by SLiM. 
+    Returns the next SLiM mutation ID for this tree sequence. This is useful 
+    because if you want to add more mutations to your SLiM tree sequence using 
+    :func:`msprime.sim_mutations`, you may need to specify the parameter 
+    `next_id` in your :class:`msprime.SLiMMutationModel` to be larger than any 
+    existing mutation IDs. Setting `next_id` equal to the output of this 
+    function will allow the mutated tree sequence to be read in by SLiM. Note
+    that this function does not work with non-SLiM tree sequences.
     '''
     max_id = -1
     for mut in ts.mutations():
         for d in mut.derived_state.split(","):
-            max_id = max(max_id, int(d))
-    return(max_id)
+            try:
+                max_id = max(max_id, int(d))
+            except ValueError:
+                raise ValueError("The derived states of mutations in the tree sequence need to be coercible to int. This is not a valied SLiM tree sequence.")
+    return(max_id+1)
 
 
 def _annotate_nodes_individuals(tables, age):
--- a/pyslim/methods.py
+++ b/pyslim/methods.py
@@ -608,6 +608,21 @@ def annotate_tables(tables, model_type, tick, cycle=None, stage="early", referen
     if reference_sequence is not None:
         tables.reference_sequence.data = reference_sequence
 
+def max_slim_mutation_id(ts):
+    '''
+    Returns the largest SLiM mutation ID in this tree sequence. This is useful because
+    if you want to add more mutations to your tree sequence using :func:`msprime.sim_mutations`,
+    you may need to specify the parameter `next_id` in your 
+    :class:`msprime.SLiMMutationModel` to be larger than any existing mutation IDs.
+    Setting `next_id` equal to the output of this function + 1
+    will allow the mutated tree sequence to be read in by SLiM. 
+    '''
+    max_id = -1
+    for mut in ts.mutations():
+        for d in mut.derived_state.split(","):
+            max_id = max(max_id, int(d))
+    return(max_id)
+
 
 def _annotate_nodes_individuals(tables, age):
     '''
--- a/pyslim/methods.py
+++ b/pyslim/methods.py
@@ -615,8 +615,11 @@ def next_slim_mutation_id(ts):
     :func:`msprime.sim_mutations`, you may need to specify the parameter 
     `next_id` in your :class:`msprime.SLiMMutationModel` to be larger than any 
     existing mutation IDs. Setting `next_id` equal to the output of this 
-    function will allow the mutated tree sequence to be read in by SLiM. Note
-    that this function does not work with non-SLiM tree sequences.
+    function will allow the mutated tree sequence to be read in by SLiM.
+    To do this, recall that the "derived state" of SLiM's mutations are
+    comma-separated strings of mutation IDs; this function just parses all derived
+    states and returns one larger than the largest integer found. It will return an error
+    if it encounters derived states that are not comma-separated strings of integers.
     '''
     max_id = -1
     for mut in ts.mutations():
--- a/pyslim/methods.py
+++ b/pyslim/methods.py
@@ -625,7 +625,7 @@ def next_slim_mutation_id(ts):
                 max_id = max(max_id, int(d))
             except ValueError:
                 raise ValueError("The derived states of mutations in the tree sequence need to be coercible to int. This is not a valied SLiM tree sequence.")
-    return(max_id+1)
+    return max_id + 1
 
 
 def _annotate_nodes_individuals(tables, age):
--- a/pyslim/methods.py
+++ b/pyslim/methods.py
@@ -621,11 +621,11 @@ def next_slim_mutation_id(ts):
     states and returns one larger than the largest integer found. It will return an error
     if it encounters derived states that are not comma-separated strings of integers.
     '''
-    max_id = -1
+    max_id = 0
     for mut in ts.mutations():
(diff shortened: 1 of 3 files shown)


pyslim/methods.py
```diff
@@ -608,6 +608,20 @@ def annotate_tables(tables, model_type, tick, cycle=None, stage="early", referen
     if reference_sequence is not None:
         tables.reference_sequence.data = reference_sequence
 
+def max_slim_mutation_id(ts):
+    '''
+    Returns the largest SLiM mutation ID in this tree sequence. If you want to
+    add more mutations to your tree sequence using :func:`msprime.sim_mutations`,
+    you may need to specify the parameter `next_id` in your 
+    :class:`msprime.SLiMMutationModel`. Setting `next_id` correctly is necessary 
+    if you want your tree sequence to be readable by SLiM.
```
petrelharp, line 617:
```suggestion
    Returns the largest SLiM mutation ID in this tree sequence. This is useful because
    if you want to add more mutations to your tree sequence using :func:`msprime.sim_mutations`,
    you may need to specify the parameter `next_id` in your 
    :class:`msprime.SLiMMutationModel` to be larger than any existing mutation IDs.
    Setting `next_id` equal to the output of this function
    will allow the mutated tree sequence to be read in by SLiM. 
```

petrelharp:
Looks good; we should also have a test that this actually lets us mutate a SLiM-produced tree sequence more and load it back in to SLiM?

mufernando:
should be good to go! thanks @petrelharp

petrelharp:
Hm - thinking a bit more, I think this should be `next_slim_mutation_id( )`, which returns the max plus one. That would mean that (a) we can say "just pass this value in" instead of "this value plus one"; and (b) it reutrns 0 in the case of no mutations, instead of -1.

petrelharp:
Ah - and, we need to clarify (and, test) what this does for tree sequences whose derived states are not SLiM-produced. In the case of binary alleles (0/1), it will return 1 (or, 2 if we change to "next"); this is unavoidable but we should say in the docstring what it actually does so that people understand. If alleles are nucleotides, it'll do
```
>>> int("A")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: 'A'
```
so maybe we should try.. catch ValueError and give an informative error message?

mufernando:
I changed it to give us the next mutation ID and introduced a try...except block to catch problems with different types of derived state.

pyslim/methods.py
```diff
@@ -608,6 +608,25 @@ def annotate_tables(tables, model_type, tick, cycle=None, stage="early", referen
...
 
+def next_slim_mutation_id(ts):
+    '''
+    Returns the next SLiM mutation ID for this tree sequence. This is useful 
+    because if you want to add more mutations to your SLiM tree sequence using 
+    :func:`msprime.sim_mutations`, you may need to specify the parameter 
+    `next_id` in your :class:`msprime.SLiMMutationModel` to be larger than any 
+    existing mutation IDs. Setting `next_id` equal to the output of this 
+    function will allow the mutated tree sequence to be read in by SLiM. Note
+    that this function does not work with non-SLiM tree sequences.
```
petrelharp, line 619:
```suggestion
    function will allow the mutated tree sequence to be read in by SLiM.
    To do this, recall that the "derived state" of SLiM's mutations are
    comma-separated strings of mutation IDs; this function just parses all derived
    states and returns one larger than the largest integer found. It will return an error
    if it encounters derived states that are not comma-separated strings of integers.
```

pyslim/methods.py
```diff
@@ -608,6 +608,25 @@ def annotate_tables(tables, model_type, tick, cycle=None, stage="early", referen
...
+    '''
+    max_id = -1
+    for mut in ts.mutations():
+        for d in mut.derived_state.split(","):
+            try:
+                max_id = max(max_id, int(d))
+            except ValueError:
+                raise ValueError("The derived states of mutations in the tree sequence need to be coercible to int. This is not a valied SLiM tree sequence.")
```
petrelharp, line 627:
```suggestion
                raise ValueError("The derived states of mutations in the tree sequence "
                                           "need to be coercible to int. This is not a valid SLiM tree sequence.")
```

pyslim/methods.py
```diff
@@ -608,6 +608,25 @@ def annotate_tables(tables, model_type, tick, cycle=None, stage="early", referen
...
+    return(max_id+1)
```
petrelharp, line 628:
```suggestion
    return max_id + 1
```

petrelharp:
I like it. A few suggested changes, and we need a test for the error.

tests/test_tree_sequence.py
```diff
@@ -32,6 +32,58 @@ def test_slim_time(self, recipe):
...
+                mts,
+                "restart_nucleotides_WF.slim",
+                tmp_path,
+                WF=False,
+        )
+        # nothing should change
+        assert pyslim.next_slim_mutation_id(mts) == pyslim.next_slim_mutation_id(rrts)
+        assert rrts.num_mutations == mts.num_mutations
+
+        def test_invalid_derived_state(self):
```
mufernando, line 74: we have a test for the error here @petrelharp

tests/test_tree_sequence.py
```diff
@@ -32,6 +32,58 @@ def test_slim_time(self, recipe):
...
+                    4,
+                    sequence_length=10,
+                    population_size=10,
+                    random_seed=10,
+            )
+            mts = msprime.sim_mutations(ts,
+                    model="jc69",
+                    rate=0.5,
+                    random_seed=23)
+            with pytest.raises(ValueError):
```
petrelharp, line 85:
```suggestion
            with pytest.raises(ValueError, match="need to be coercible to int"):
```

pyslim/methods.py
```diff
@@ -608,6 +608,29 @@ def annotate_tables(tables, model_type, tick, cycle=None, stage="early", referen
...
+    if it encounters derived states that are not comma-separated strings of integers.
+    '''
+    max_id = -1
+    for mut in ts.mutations():
+        for d in mut.derived_state.split(","):
+            try:
+                max_id = max(max_id, int(d))
+            except ValueError:
+                raise ValueError("The derived states of mutations in the tree sequence "
+                                           "need to be coercible to int. This is not a valid SLiM tree sequence.")
```
petrelharp, line 631:
```suggestion
                raise ValueError("The derived states of mutations in the tree sequence "
                                           "need to be coercible to int. This is not a valid SLiM tree sequence.")
```

petrelharp: maybe??? check for tabs?

mufernando:
this should be good to go now!

tests/test_tree_sequence.py
```diff
@@ -32,6 +32,58 @@ def test_slim_time(self, recipe):
...
+    '''
+    Tests for the function that returns the largest SLiM mutation ID.
+    '''
+    @pytest.mark.parametrize('recipe', [next(recipe_eq())], indirect=True)
+    def test_next_id(self, recipe):
+        ts = recipe["ts"]
+        mt_ids_str = ','.join(tskit.unpack_strings(ts.tables.mutations.derived_state, ts.tables.mutations.derived_state_offset))
+        mt_ids = [int(i) for i in mt_ids_str.split(',')]
+        max_mt_id = max(mt_ids)
+        assert max_mt_id+1 == pyslim.next_slim_mutation_id(ts)
```
petrelharp, line 45:
```suggestion
        assert max_mt_id + 1 == pyslim.next_slim_mutation_id(ts)
```

tests/test_tree_sequence.py
```diff
@@ -32,6 +32,58 @@ def test_slim_time(self, recipe):
                 mut_time = max([x['slim_time'] for x in mut.metadata['mutation_list']])
                 assert mut_time == pyslim.slim_time(ts, mut.time, stage="early")
 
+class TestNextMutationID(tests.PyslimTestCase):
+    '''
+    Tests for the function that returns the largest SLiM mutation ID.
+    '''
+    @pytest.mark.parametrize('recipe', [next(recipe_eq())], indirect=True)
```
petrelharp, line 39: we want to test this on lots of recipes (like, all of them?) not just the first one (which might have no mutations!)
```suggestion
```

petrelharp:
Whoops - testing in more situations has turned up an issue: a derived or ancestral state of `""` is valid (it means "no mutations"), and this produces an error. I think that right now the main implementation of this function would wrongly throw an error in this case, in fact.

petrelharp:
Huh, and actually, this method should also iterate over `sites.ancestral_state`.

pyslim/methods.py
```diff
@@ -608,6 +608,30 @@ def annotate_tables(tables, model_type, tick, cycle=None, stage="early", referen
...
+    '''
+    max_id = 0
+    for mut in ts.mutations():
+        for d in mut.derived_state.split(","):
+            try:
+                max_id = max(max_id, int(d or 0)) # the or zero will allow empty strings to be converted
+            except ValueError:
+                raise ValueError("The derived states of mutations in the tree "
+                                 "sequence need to be coercible to int. This "
+                                 "is not a valid SLiM tree sequence.")
```
petrelharp, line 632: Gee, that's clever. Hm, but the only time we'd get an empty string is if the derived state is `""` (so e.g., `"9,"` is not valid). So, how about this, which will correctly not work with `"9,"` and is a bit more clear that it's right, also?
```suggestion
        ds = mut.derived_state
        if len(ds) > 0:
            for d in mut.derived_state.split(","):
                try:
                    max_id = max(max_id, int(d))
                except ValueError:
                    raise ValueError("The derived states of mutations in the tree "
                                     "sequence need to be coercible to int. This "
                                     "is not a valid SLiM tree sequence.")
```

mufernando:
@petrelharp and I talked about the ancestral state column in the Sites Table today and we don't see how that could be used here? SLiM doesn't use that column yet.

PR #291, opened 2022-08-12
added slim clone and compile to GH actions

andrewkern:
set up GH actions to build `SLiM` and test `pyslim` on windows

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -93,7 +93,7 @@ jobs:
       #    cd SLiM/Release
       #    cmake -D CMAKE_BUILD_TYPE=Release ..
       #    make -j 2
-
+      #
       - name: Run tests
         run: |
           source ~/.profile
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -83,7 +83,6 @@ jobs:
         run: |
           mkdir -p /usr/local/miniconda/envs
           sudo cp -r ~/osx-conda /usr/local/miniconda/envs/anaconda-client-env
-
       #- name: Build SLiM
       #  if: steps.cache.outputs.cache-hit != 'true' &&
       #    (matrix.os == 'macos-latest' || matrix.os == 'ubuntu-latest')
@@ -93,12 +92,11 @@ jobs:
       #    cd SLiM/Release
       #    cmake -D CMAKE_BUILD_TYPE=Release ..
       #    make -j 2
-      #
       - name: Run tests
         run: |
           source ~/.profile
           conda activate anaconda-client-env
-       #   export PATH=$PWD/SLiM/Release:$PATH # don't need to add SLiM to path, we ain't building it
+          # export PATH=$PWD/SLiM/Release:$PATH # don't need to add SLiM to path, we ain't building it
           which slim
           slim -v
           pytest -xv -n2 tests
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -3,17 +3,16 @@ name: Tests
 on:
   pull_request:
   push:
-    branches: [main]
+    branches: [main, build_slim_gh_Actions]
```
andrewkern, line 6: we should probably remove this other branch push action... was using this to develop

petrelharp: ah, yep. should do.

andrewkern: done

tests/conftest.py
```diff
@@ -192,6 +196,9 @@ def run_slim_restart(cls, in_ts, recipe, out_dir, subpop_map=None, **kwargs):
         # on it, saving to files in out_dir.
         infile = os.path.join(out_dir, "in.trees")
         in_ts.dump(infile)
+        # for happy windows filepaths       
+        infile_str = infile.replace('\\', '\\\\')
+        print(infile_str)
```
petrelharp, line 201: stray print

.github/workflows/tests.yml
```diff
@@ -85,15 +84,15 @@ jobs:
...
-      #     make -j 2
+      - name: Build SLiM
+        if: steps.cache.outputs.cache-hit != 'true' &&
+          (matrix.os == 'macos-latest' || matrix.os == 'ubuntu-latest')
+        run: |
+          git clone https://github.com/messerlab/SLiM.git
+          mkdir -p SLiM/Release
+          cd SLiM/Release
+          cmake -D CMAKE_BUILD_TYPE=Release ..
+          make -j 2
```
petrelharp, line 95: what's up with these changes?

tests/conftest.py
andrewkern: actually those prints are useful! they only appear if you use `pytest -s`

happy to delete ofc

.github/workflows/tests.yml
andrewkern: just a bit of a noisy diff it looks like from uncommenting that code block so we build on macos / ubuntu, but guess we don't need it?

tests/conftest.py
andrewkern: okay got rid of this

.github/workflows/tests.yml
andrewkern: commenting these out

petrelharp:
Sorry, yet another question - I see that `ci-environment.yml` is used only for Windows, while various `requirements.txt` serve for the other platforms. Is there a good reason for this, or can I just make them all use the same file?

andrewkern:
right. so i'm doing things a bit different in the Windows section than what you had originally. I'm using conda with a yaml environment file on instantiation that builds the virtual environment with everything preinstalled. the way you had been doing it was to create a conda environment, and then in separate steps install the conda reqs and then the pip reqs. Using a single conda env file on instantiations collapses all these steps down to one step and saves a bunch of time on the GH action workflow.

I didn't edit the macos / ubuntu side of things as it currently just works. IMO the right thing to do next would be to edit those sections of the workflow to do similar, and then to ditch conda and move to a provisioned micromamba setup. I'd vote we do that on a separate PR, once this is merged as an example.

andrewkern:
closing this cause i've got a cleaner PR on deck

PR #299, opened 2022-09-12
use metadata vector

petrelharp:
Over in https://github.com/popsim-consortium/stdpopsim/issues/1334 we found out that `convert_alleles` is taking way too long. I had a try at changing it to use `metadata_vector` (in this PR); however, there's no appreciable change in speed. (However, note that at least I noticed that the complicated stuff I was doing had no effect - either a bug or SLiM does always put the most recent mutation first.)


bhaller:
> ...either a bug or SLiM does always put the most recent mutation first.)

I'm not sure whether that is guaranteed.  It might be different for new mutations added by SLiM itself versus those added by addNewMutation() / addNewDrawnMutation(), and I'm not sure how removeMutations() might affect things.  This could of course be checked, but I don't have time to get into it right at the moment.  Might be easiest to determine empirically, rather than by looking at the code.  :->  But then of course there is also the question of whether SLiM _guarantees_ any particular behavior, or whether whatever it does right now could change in the future without a "break of contract".  I would lean towards the latter, especially since neither of us actually knows what the "contract" might be, and it might be inconsistent, and so forth.  I don't think there's any "contract" in this area.

petrelharp:
Codecov hasn't run for some reason (I've just added the action to github), but I've run it locally, and it looks good.

pyslim/methods.py
```diff
@@ -208,8 +204,13 @@ def generate_nucleotides(ts, reference_sequence=None, keep=True, seed=None):
     tables.mutations.clear()
     sets = [[k for k in range(4) if k != i] for i in range(4)]
     states = np.full((ts.num_mutations,), -1)
+    k = tables.sites.position.astype('int')
+    aa_list = np.searchsorted(
+            NUCLEOTIDES,
+            np.frombuffer(tables.reference_sequence.data.encode('utf-8'), dtype='S1')[k],
```
benjeffery, line 210: I think this should be dtype='U1'

pyslim/methods.py
```diff
@@ -138,21 +140,15 @@ def convert_alleles(ts):
...
     if np.any(nuc_inds == -1):
         raise ValueError("All mutations must be nucleotide mutations.")
-    if not has_refseq:
-        raise ValueError("Tree sequence must have a valid reference sequence.")
-    aa = [ts.reference_sequence.data[k] for k in tables.sites.position.astype('int')]
     da = np.array(NUCLEOTIDES)[nuc_inds]
-    tables.sites.packset_ancestral_state(aa)
     tables.mutations.packset_derived_state(da)
+    k = tables.sites.position.astype('int')
+    aa = np.frombuffer(ts.reference_sequence.data.encode('utf-8'), dtype='S1')[k]
```
benjeffery, line 150: I think this should be dtype='U1'

benjeffery (commented):
I know I originally suggested `S1`, but I think `U1` might be more correct.

PR #301, opened 2022-09-15
updates to demo use of meanParentAge

silastittes:
Edits to use the new `meanParentAge` property introduced in SLiM 4.0.1, with minor edits in body of text to explain the update.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/cache@v2
         env:
           # Increase this to reset the cache if the key hasn't changed.
-          CACHE_NUM: 1
+          CACHE_NUM: 2
         with:
           path: |
             c:\Miniconda\envs\anaconda-client-env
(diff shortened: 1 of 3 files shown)


petrelharp:
This is great, Silas. I was hoping you might stick in the comparison of site-diversity to branch-length-diversity here as well somehow, with some explanation? (before the 'recapitation' bit)  That would make things more concrete?

petrelharp:
(Also see the errors in building the docs.)

silastittes:
DOH! I think I just forgot the `git add` the new .trees file.  
But thanks! And I agree adding the branch length comparisons makes sense, will make those updates soon!

silastittes:
I updated the cache value and added some site-diversity to branch-length-diversity details. Hopefully looks good and  build docs works! <fingers crossed>

petrelharp:
Weird I have to keep clicking "approve" on your checks - I will look into this later.

petrelharp:
Great! I put in a bit more explanation (e.g., something about *why* that check works), and removed the need to re-simulate without mutations - what do you think?

silastittes:
Awesome! Looks great and helped to clarify several ideas I was unsure about.

PR #319, opened 2023-07-17
explain no adjustment needed for WF sims

petrelharp:
From the email list:

> I am unsure whether I need to adjust the mutation, recombination rate and Ne by the generation time when recapitating with pyslim and adding neutral mutations with msprime. I have read many times the section about time units mismatch in pyslim manual (https://tskit.dev/pyslim/docs/latest/time_units.html), but most of the examples are nonWF oriented.

### DIFF
--- a/docs/time_units.md
+++ b/docs/time_units.md
@@ -75,6 +75,10 @@ then you'd want the mutation rate in units of mutations per year.
 
 ## Coalescence, recombination, and migration rates in recapitation
 
+*Note: this section is relevant for nonWF SLiM simulations:
+in a single-species WF simulation, one tick = one generation,
+so no adjustment is needed.*
+
 If you recapitate with msprime, then the same issue occurs for coalescence rates.
 This is less obvious, because "coalescence rate" is not an argument to msprime;
 rather, the coalescence rate in a given population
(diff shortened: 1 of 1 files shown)


docs/time_units.md
```diff
@@ -75,6 +75,10 @@ then you'd want the mutation rate in units of mutations per year.
 
 ## Coalescence, recombination, and migration rates in recapitation
 
+*Note: this section is relevant for nonWF SLiM simulations:
```
bhaller, line 78: maybe "is only relevant for"?

docs/time_units.md
```diff
@@ -75,6 +75,10 @@ then you'd want the mutation rate in units of mutations per year.
...
+in a single-species WF simulation, one tick = one generation,
```
bhaller, line 79: good catch on "single-species"; I guess multispecies models that use different timescales would need adjustment; do you intend to add discussion of that, or do you figure users doing that are advanced enough that they can infer what needs to be done?

petrelharp: I don't want to mention that here, no. =)

docs/time_units.md
petrelharp:
```suggestion
*Note: this section is only relevant for nonWF SLiM simulations:
```

Issue #317, opened 2023-06-26
test on annotating tsinfer'ed tree sequence


petrelharp:
... there may be problems with individuals having `-1` populations

PR #328, opened 2023-08-23
typo fix metdadata

Tatianabellagio:
Fixed a little typo in the code: changed "ind.metdata" to "ind.metadata" that was making the function crash when annotating the individuals table


petrelharp:
Whoops - thanks for the fix! (And, #317 is looking more urgent.)

PR #332, opened 2023-10-11
Suggestion for vignette: a spatial simulation

jiseonmin:
Hey, thank you for a helpful vignette on spatial simulations! I saw that the third entry of the location list is included when calculating the geographic distance. May I suggest using only the first two entries, since the example is in a two dimensional space?


petrelharp:
OMG good catch. (And I somehow just noticed these PRs.)

PR #335, opened 2023-10-30
fix navigation with keys warning

petrelharp:
This does as https://github.com/tskit-dev/msprime/pull/1395 and should help with https://github.com/tskit-dev/tskit-site/pull/192 ?


benjeffery:
Thanks @petrelharp - Sadly the site will need some extra work as it builds stable versions that we'll have to patch! Will get it fixed though.

PR #344, opened 2024-06-10
Fix mutation rate in time units example

lkirk:
The branch diversity stat was being scaled by 1e-8 instead of 2e-8, leading to the underestimation of branch diversity when scaled by generation time.

I've corrected the code example and the text.


petrelharp:
Good catch - and thanks for the PR! [Here's](https://github.com/tskit-dev/pyslim/blob/4cb44e23608c33ed0086509674970adeb182cf7c/docs/generation_time.slim#L5) where mutation rate is indeed set to 2e-8.
(The test failure is a transient CI failure; I'll merge this.)

PR #396, opened 2025-09-09
Add reusable lint workflow

benjeffery:


benjeffery:
Just realised this repo doesn't have any linting. This is what it would look like to add - there are many issues to fix to make the lint pass. In the other tskit repos I have sometimes found real issues with this so might be worth doing, but I don't have time right now.

PR #424, opened 2026-05-21
add a bit on names in vcf file

petrelharp:
In response to a question of how to get more informative names in the VCF.


petrelharp:
This is just a docfix; the test failure is unrelated and recorded in #425.

PR #427, opened 2026-06-17
fix codecov action

petrelharp:
See https://github.com/tskit-dev/tskit/pull/3467


petrelharp:
Failing, but for other reasons.

PR #434, opened 2026-08-19
bug fixes in add_mutation_metadata_tables

jeffspence:
I've been trying to use `pyslim` in the [multi-trait addition to stdpopsim](https://github.com/jeffspence/stdpopsim/tree/traits_into_slim_engine) and was getting some errors related to `pyslim.add_mutation_metadata_tables` .  Digging into it, there were two separate things going on:

1. If the `tables` object passed to `pyslim.add_mutation_metadata_tables` has multiple traits, that information is not passed into `default_slim_metadata()`, which then results in errors resulting from mismatching dimensions.  Adding `num_traits=len(ts_metadata["SLiM"]["traits"])` as a kwarg to `default_slim_metadata()` fixes that.
2. if `mut_ids` is empty, then `np.array(mut_ids, dtype="int")` only has one dimension (of length 0), but subsequent code assumes that it is 2 dimensional.  Wrapping the code to extend `ts_metadata["SLiM_mutation_list"]` in an if block deals with this, and shouldn't affect anything because if `mut_ids` is empty, then there's nothing to add to `ts_metadata["SLiM_mutation_list"]` anyway.

I didn't write any tests for these because I'm not sure where to start (and a bunch of the tests are currently failing independent of the code changes proposed in this PR) but I'd be happy to do that with a bit of guidance.

Edit: I also found a missing word in the `convert_alleles` docstring.


petrelharp:
Whoops, sorry - I seem to have missed this? I'll get this in.

petrelharp:
Hey @jeffspence - can you let me know what was the code that was wrong that you needed to fix with this?

jeffspence:
@petrelharp I triggered the errors in the fork of stdpopsim where I'm working on traits stuff (https://github.com/jeffspence/stdpopsim/tree/traits_into_slim_engine).  Running the tests in `TestTraits` in `test_slim_engine.py` with the non-patched version of pyslim incorporating the above changes results in a bunch of errors.

A MWE that shows both of the issues is in the attached files.  [mwe.zip](https://github.com/user-attachments/files/31667136/mwe.zip) contains `mwe.slim` which initializes two traits and then simulates a tree sequence (present in `mwe.trees`).  The code in `mwe.py` essentially tries to recapitate it and then add mutations.  If the mutation rate is low, no mutations get added and the error in number 2 [here](https://github.com/tskit-dev/pyslim/pull/434#issue-5197998302) gets triggered.  If the mutation rate is high enough that some mutations get added, then error number 1 gets triggered.

Let me know if any of that is unclear, or if I'm doing something trivially wrong.

petrelharp:
Thanks for this! I've got some tests in that should trigger these issues. I guess we need some more traits-related tests in here!

PR #442, opened 2026-08-26
metadata updates

petrelharp:
Now:
- mutations use metadata once again
- individuals no longer have a `flags`


petrelharp:
Windows tests are failing; I'm merging and dealing with that later.

PR #449, opened 2026-09-03
test for trait consistency

petrelharp:
Not working yet!


petrelharp:
Hey, @bhaller - I'm going to merge this, but could you have a look at the doc changes? They should be built at https://tskit.dev/pyslim/docs/latest tomorrow; there's some edits to the "traits" page, including a "technical details" bit at the bottom.

bhaller:
> Hey, @bhaller - I'm going to merge this, but could you have a look at the doc changes? They should be built at https://tskit.dev/pyslim/docs/latest tomorrow; there's some edits to the "traits" page, including a "technical details" bit at the bottom.

Sure, reviewing here:

- "for an trait in SLiM" -> "a trait"
- "start off with this simulation" -> "start off with the simulation below" (since there's some text first, so the referent for "this simulation" might be unclear)
- I don't understand why you do "m2.convertToSubstitution = F"; you have a comment "necessary to disentagle from baseline offset", but the baseline offset is a constant and is persisted in the top-level metadata, so what is the problem?

Trait distributions section:

- "We can use the helpful metadata_vector method..." – it'd be helpful to show what this gives you as a result (in terms of the structure of the data); it's not clear to me what this is doing.  I think `j` is the trait index?  If so, it'd be helpful to name that variable "trait_index" instead of just `j`.  :->
- "we can get the genetic contributions by subtracting the individual-level offsets" – hmm, I wouldn't put it this way.  You're leaving the baseline offsets in, and it is not clear that those represent "genetic contributions" in general.  Maybe just say "we can subtract out the environmental noise represented by the individual-level offsets to get the phenotypes of the individuals without that noise" or some such?  Or – do you even really need to subtract out those offsets at all?  Your plot axes are labeled "phenotype1" and "phenotype2", and the individual offsets are part of the "phenotype".  The point clouds will just be a bit fuzzier, which seems like it would be fine.  Oh, I see, that is in fact what you did – it's confusing because you show getting the genetic contributions by subtracting offsets, but then you plot the phenotypes including the offsets, not using what you just calculated.  Maybe move the subtraction of the offsets to be down below the plot, doing things one step at a time in a more linear order?
- "moving from its initial condition at (-5, 5) towards" – it's a bit unfortunate that the first snapshot is at tick 5000, not tick 1.  It'd be nice to adjust the timing of the snaps so you get to see the initial position in the plot.  (BTW, do you mean "position" instead of "condition"?  I guess condition sorta works too, but sounds a bit odd to me.)  I kinda regret that I made ticks 1-based instead of zero, based, it is a constant source of annoyance with this kind of thing since obviously it is nicer to plot at intervals where % 5000 == 0 rather than % 5000 == 1.  Actually, there's nothing _really_ preventing me from relaxing that and allowing tick 0 to be used, I guess...?  Right now `initialize()` time is considered tick 0, internally, but I don't think that is really user-visible, so I could shift that to tick -1 and allow tick 0 to be used.  Do you think that would be a good idea, or would it cause you new nightmares in the time conversions in pyslim?  Anyhow, just a thought since this problem keeps coming up.
- Are you sure you want to use alpha=0.5?  It puzzled me for a little while, because it looks (in the final plot in this section, showing the phenotypes minus the offsets) like there are more colors than there are time points, and some of the colors – notably the grayish point at about x=17.5 – don't appear to be in the color scale shown.  Then I realized that that's due to the alpha, and to overplotting of many points in some places and just one or a few in other places.  I know in the previous plot you want the cloud to be more visible as a bunch of discrete points; maybe changing the outline color of the points instead of using alpha would be clearer?  Anyhow, a nit, maybe I was just slow to understand what I was seeing.  :->

Mutation effects section:

- Adding a mention that the model is neutral, and just drifts for 1000 ticks, would be clarifying.  It might also be good to mention that since the traits have no effect on fitness, their values won't get calculated by SLiM at all except in the final tick when you call demandPhenotype(); it doesn't bother calculating things it doesn't need and wasn't asked for.
- You might be interested to know that under the hood, optimization for independent dominance happens at the trait level.  In this model each trait is influenced by both m1 and m2 mutations, and m2 is ind-dom but m1 is not, so neither trait gets and ind-dom optimization; the performance will be the same as if you used a default dominance coefficient of 0.5 instead of NAN.  It's still good to use NAN, because why not – it expresses the hopeful intent that optimization might occur.  But it won't.  The reasons for this get WAY down into the weeds of how ind-dom works internally, not worth getting into here; but I thought you might just like to know.  :->
- "For example, here’s the first mutation" – this isn't really an "example" what you just happened previously; you're not using ts_metadata or mut_metadata.  So maybe just say "Here's the first mutation"?
- "SLiM mutaiton IDs" typo
- "in the mutation’s “derived state”" – maybe at this point it would be helpful to point out that the derived state info is in two places, one as ASCII and one as binary, etc?  I know it gets into the weeds a little, but I imagine the reader will be wondering, if they're paying attention.
- Maybe it is a bit confusing have mut.metadata and mut_metadata?  I've been trying to refer to what you've named mut_metadata as the "mutation metadata table", versus referring to mut.metadata as the mutation table's metadata column.  Maybe rename mut_metadata to mut_metadata_table?
- for `def additive_effect` and `def additive_genetic_value` it seems a bit unfortunate to need to pass in `num_traits`; and although it is convenient for your example code to have the default be 2, it seems like in general one-trait models will probably be more common than two-trait models, so maybe the default should be 1?  But can the number of traits be interred from somewhere?  It has `mut_metadata`, so it could get the number of traits from the structure of that metadata (the number of per-trait entries), perhaps?
- These functions seem maybe pretty useful; should functionality like this be build into pyslim?
- "and here we do nothing with the substitutionOffset" needs grammar fix, delete "and" probably

Multiplicative and logistic traits section

- "we’ll change the first trait to be multiplicative" – note that for m1, a mutational effect could be negative since the effect size is drawn from a normal distribution.  Not common, since the SD is only 0.01 (and so 1+s or 1+hs is quite unlikely to be negative).  But I don't see you handling that possibility in your code that follows, so your code is kinda relying on that to be true; if the SD of the DES for m1 were changed to be much broader, I think (?) your code would produce an incorrect result.  IIRC, SLiM clamps each separate mutational effect to a minimum of 0.0.  (And it raises if a baseline offset for a multiplicative trait is set to < 0.0, since that is just clearly wrong.)  You might want to do the same?
- "Happily, these again agree, up to floating point error" – a bit off-topic, but I'm curious.  It looks like SLiM's values are double-precision floats but pyslim's values are single-precision floats, just based upon the number of digits printed.  Is that true?  Does python default to single-precision float unless instructed otherwise, or something?
- "and cumulative mutation effects" add "from" (i.e., "and from") for parallel construction.
- "The contribution of 0, 1, or 2 copies of a SLiM mutation to a diploid" – for completeness it would be good, in this paragraph, to also mention the haploid case, which is 0, 2a or 1, 1+s – no dominance coefficient at all.
- "is either 0, 2hs, or 2s (additive)" – I always refer to the effect size as `a` for additive traits, not `s`, but I realize that having two different symbols is clunky.
- "(This is because genome structure is set up before .trees files can be loaded.)"  I'm not sure what you mean by this...?  We certainly could write out that kind of stuff to metadata, we just don't.  :->  (LMK if you want me to!)
- "The potentially missing information is in the script: are these mutation types converting to substitutions or not?"  True.  It might be clarifying to point that using convToSub=T together with subAcc=F is, in general, a dangerous combination that should probably not be used.  I recently added a warning in SLiM about it, in fact:

> #WARNING (Species::DoSubstitutionAccumulation): a non-neutral mutation was converted to a substitution with substitution accumulation disabled, either (1) in a nonWF model, or (2) for a trait that does not have a direct effect on fitness.  This specific situation typically produces an undesirable, non-biological 'jump' in trait value, because the mutation's effect on the trait is lost when the mutation is turned into a substitution.  (This is not generally a problem for WF models with a trait that has a direct effect on fitness, because fitness is relative in the WF model.)  To fix this warning, you would typically want to either turn substitution off for this mutation type (with convertToSubstitution=F), or turn substitution accumulation on for the trait being affected (with substitutionAccumulation=T).

(It's a long and clunky warning, because an exception is carved out for WF models with traits that have a direct effect on fitness – i.e., SLiM's "default trait", for backward compatibility.  But apart from that carve-out, you get the warning.  So the problematic case that you're talking about here is _also_ a problem in SLiM, and is generally frowned upon.)

- "differing effects of fixed effects on individuals" maybe instead "differing effects of mutations on individuals", and add the word "hemizygosity" somewhere in the sentence since that is the key concept?

Overall it seems great; very clear, and really sets out the logic of how things work.  It does focus a lot – almost to the exclusion of other considerations – on matching SLiM's phenotype calculations with equivalent calculations in python.  That's great for building conceptual clarity, but in practice it is probably not something people will do much, right?  Since the trait values are already in the metadata for them to use.  Are there other aspects of working with trait-based data in python that this section ought to also talk about, since they will be likely to come up real-world usage?  Maybe things like calculating V_A, V_D, V_G, V_P on the python side, or other quant-gen kinds of analyses?  Or something with running a PCA based upon multi-trait phenotype data, or some such?

And I'm also wondering: is there anything that should be said about tstrait here?  Is there any useful way to interface tstrait with SLiM's traits?  Does tstrait have any useful analysis tools that can somehow be brought to bear upon data from SLiM simulations?  (Maybe the answers here are simply "no"; I don't know much about tstrait.)

OK, I hope the above is helpful.  Good stuff!  :->

petrelharp:
>  Right now initialize() time is considered tick 0, internally, but I don't think that is really user-visible, so I could shift that to tick -1 and allow tick 0 to be used. Do you think that would be a good idea, or would it cause you new nightmares in the time conversions in pyslim? Anyhow, just a thought since this problem keeps coming up.

I'm sympathetic, but I think "no". (It wouldn't cause problems with pyslim, but I could imagine it causing unforseen problems with people's scripts?)

bhaller:
> > Right now initialize() time is considered tick 0, internally, but I don't think that is really user-visible, so I could shift that to tick -1 and allow tick 0 to be used. Do you think that would be a good idea, or would it cause you new nightmares in the time conversions in pyslim? Anyhow, just a thought since this problem keeps coming up.
> 
> I'm sympathetic, but I think "no". (It wouldn't cause problems with pyslim, but I could imagine it causing unforseen problems with people's scripts?)

Well, all existing scripts would work the same.  If your script says "1 early()" for the initial event, it'd still run in tick 1.  I'm just proposing that "0 early()" become legal.  But I agree that there are very possibly non-obvious side effects here, nevertheless.

petrelharp:
> These functions seem maybe pretty useful; should functionality like this be build into pyslim?

They do seem useful! And there's even more useful versions of them in the [unit tests](https://github.com/tskit-dev/pyslim/blob/eca7d207acb578c7abfd324765d20be76d76b8cc/tests/test_metadata.py#L432), but I don't want to lock into providing that functionality without a clear use case.

petrelharp:
> IIRC, SLiM clamps each separate mutational effect to a minimum of 0.0.

Ah, I took that to mean that if one set a mutational effect to less than 0.0 the effect got set to 0.0, rather than when it got used it'd be 0.0.

petrelharp:
> "Happily, these again agree, up to floating point error" – a bit off-topic, but I'm curious. It looks like SLiM's values are double-precision floats but pyslim's values are single-precision floats, just based upon the number of digits printed. Is that true? Does python default to single-precision float unless instructed otherwise, or something?

I've checked; python is (on my machine anyhow) using double-precision floats, I think there was just a difference in how these were being printed out.

petrelharp:
> "(This is because genome structure is set up before .trees files can be loaded.)" I'm not sure what you mean by this...? We certainly could write out that kind of stuff to metadata, we just don't. :-> (LMK if you want me to!)

This is a bigger topic! But I think this is the fundamental reason we don't write out things like this to the metadata already: genome set-up is required (?) to have happened before the point at which you can load a tree sequence? Maybe it'd be okay to write out this particular thing? Right now we don't write out anything about mutation types, and correct me if I"m wrong but it sounds like a giant pain to change things over for the set of mutation types to be dynamic.

In short: I think it's fine as-is - could be cleaner, but fine.

petrelharp:
>  It does focus a lot – almost to the exclusion of other considerations – on matching SLiM's phenotype calculations with equivalent calculations in python. That's great for building conceptual clarity, but in practice it is probably not something people will do much, right? 

Yes, I hope so! That's why this is down in the "Technical details" section.

Good question about tstrait. And, good suggestion about calculating V_A. I meant to do that but got sidetracked with the technical details. =)

Okay, i've got all these edits off in another PR!

bhaller:
> > IIRC, SLiM clamps each separate mutational effect to a minimum of 0.0.
> 
> Ah, I took that to mean that if one set a mutational effect to less than 0.0 the effect got set to 0.0, rather than when it got used it'd be 0.0.

Yeah; you can't clamp the effect size of the mutation itself, because there are two (or three, with hemizygosity) possible effects based on that effect size – 1+s and 1+hs.  Even if 1+s is less than zero, 1+hs might not be (or vice versa, for h outside of the usual [0,1] range).  So you have to calculate the effect and clamp that, not the effect size.  (SLiM actually pre-calculates clamped effects for 1+s and 1+hs, to avoid having to do that work over and over.)

bhaller:
> > "(This is because genome structure is set up before .trees files can be loaded.)" I'm not sure what you mean by this...? We certainly could write out that kind of stuff to metadata, we just don't. :-> (LMK if you want me to!)
> 
> This is a bigger topic! But I think this is the fundamental reason we don't write out things like this to the metadata already: genome set-up is required (?) to have happened before the point at which you can load a tree sequence? Maybe it'd be okay to write out this particular thing? Right now we don't write out anything about mutation types, and correct me if I"m wrong but it sounds like a giant pain to change things over for the set of mutation types to be dynamic.
> 
> In short: I think it's fine as-is - could be cleaner, but fine.

We might be talking pas each other here?  I was just suggesting that we could write out the mutation type stuff to top-level metadata, for reference on the python size – not actually _using_ it in any particular way.  Your text made it sound like we don't write it out "because genome structure is set up before .trees files can be loaded"; but we could certainly write it out, there's no obstacle to that.  We just (very definitely!) don't want to "change things over for the set of mutation types to be dynamic" <shudder>.  So I'm just suggesting a small rewording – we don't write it out because, so far, it hasn't seemed useful to do so, more than anything else.  :->

bhaller:
> > It does focus a lot – almost to the exclusion of other considerations – on matching SLiM's phenotype calculations with equivalent calculations in python. That's great for building conceptual clarity, but in practice it is probably not something people will do much, right?
> 
> Yes, I hope so! That's why this is down in the "Technical details" section.

Seems like most of the content _before_ "Technical details" is also about calculating trait values from the genetics, starting with "Now let’s see how to use mutation information to calculate genetic values directly."  That's why I was commenting that the bulk of the focus seems to be on doing that.  It's fine; I just think there are probably other topics worth talking about too.  :->

> Good question about tstrait. And, good suggestion about calculating V_A. I meant to do that but got sidetracked with the technical details. =)

:->

> Okay, i've got all these edits off in another PR!

Awesome!  Things are looking great!

